PR #22, opened 2020-01-02
DM-124 Complete Organization List Page

dhruvio:
- Continued/completed the org list page in the front-end.
- @sutherlanda Please review the change I made to `db.ts`. The issue was the vendors were not receiving all organizations from the `readMany` endpoint — they were only receiving the orgs they owned. I updated the `readManyOrganizations` function, but I'm not sure if there is a better way to do this with SQL.

### DIFF
--- a/src/back-end/lib/db.ts
+++ b/src/back-end/lib/db.ts
@@ -181,30 +181,39 @@ export async function deleteSession(connection: Connection, id: Id): Promise<nul
   return null;
 }
 
+/**
+ * Return all organizations from the database.
+ *
+ * If the user is:
+ *
+ * - An admin: Include owner information for all organizations.
+ * - A vendor: Include owner information only for owned organizations.
+ */
+
 export async function readManyOrganizations(connection: Connection, session: Session): Promise<OrganizationSlim[]> {
   const query = connection('organizations')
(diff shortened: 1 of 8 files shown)


src/back-end/lib/db.ts
```diff
@@ -181,30 +181,39 @@ export async function deleteSession(connection: Connection, id: Id): Promise<nul
...
   const results = await query;
-
-  return Promise.all(results.map(async raw => await rawOrganizationToOrganizationSlim(connection, raw)));
+  // Only include ownership information for vendors' owned organizations.
+  return Promise.all(results.map(async raw => {
+    if (session.user && session.user.type === UserType.Vendor && raw.ownerId !== session.user.id) {
+      raw = {
+        ...raw,
+        ownerId: undefined,
+        ownerNmae: undefined
```
sutherlanda: Typo here: `ownerNmae`

dhruvio: Fixed

dhruvio:
Committed some changes:

- Hide owner columns in org list page for anonymous and Government (non-admin) users.
- Hide back links on sign in and sign up pages based on feedback from Renee.

PR #31, opened 2020-01-09
DM-151 User profile page's Organizations tab

dhruvio:
Includes back-end changes and bugfixes (@sutherlanda please review back-end changes).

### DIFF
--- a/src/back-end/lib/db.ts
+++ b/src/back-end/lib/db.ts
@@ -333,7 +333,7 @@ export async function readManyAffiliations(connection: Connection, userId: Id): 
       'organizations.active': true
     });
 
-  return Promise.all(results.map(async raw => rawAffiliationToAffiliation(connection, raw)));
+  return Promise.all(results.map(async raw => rawAffiliationToAffiliationSlim(connection, raw)));
 }
 
 export async function readOneAffiliation(connection: Connection, user: Id, organization: Id): Promise<Affiliation> {
@@ -359,7 +359,8 @@ export async function readOneAffiliationById(connection: Connection, id: Id): Pr
   return result;
 }
 
-interface RawAffiliationToAffiliationParams {
+interface RawAffiliationToAffiliationSlimParams {
+  id: Id;
   user: Id;
   organization: Id;
   membershipType: MembershipType;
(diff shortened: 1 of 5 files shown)


src/back-end/lib/db.ts
```diff
@@ -351,35 +352,42 @@ export async function readOneAffiliation(connection: Connection, user: Id, organ
 
 export async function readOneAffiliationById(connection: Connection, id: Id): Promise<Affiliation> {
   const result = await connection('affiliations')
-    .where({
-      id
-    })
+    .where({ id })
+    .andWhereNot({ membershipStatus: MembershipStatus.Inactive })
```
dhruvio: @sutherlanda FYI this `andWhereNot` function call was missing in several affiliation functions in `db.ts`

dhruvio: It resulted in a bug that would vend inactive affiliations to API consumers.

src/back-end/lib/resources/affiliation.ts
```diff
@@ -182,7 +182,7 @@ const resource: Resource = {
         }
 
         const existingAffiliation = validatedAffiliationId.value;
-        if (existingAffiliation.membershipType === MembershipType.Owner && await readActiveOwnerCount(connection, request.query.organization) === 1 ) {
+        if (existingAffiliation.membershipType === MembershipType.Owner && await readActiveOwnerCount(connection, existingAffiliation.organization.id) === 1 ) {
```
dhruvio: @sutherlanda I think you forgot to change this ref from the query string to the `existingAffiliation`. Let me know if the query param ref was intentional.

src/shared/lib/resources/affiliation.ts
```diff
@@ -24,7 +24,8 @@ export interface Affiliation {
 
 // Used when returning a list of the current user's affiliations
 export interface AffiliationSlim {
-  organizationName: string;
+  id: Id;
+  organization: Pick<Organization, 'id' | 'legalName'>;
```
dhruvio: @sutherlanda FYI I expanded the `AffiliationSlim` type with necessary fields for the front-end.

src/back-end/lib/db.ts
sutherlanda: Got it, thanks for correcting.

src/back-end/lib/resources/affiliation.ts
sutherlanda: Yes that was an oversight, thank you.

PR #32, opened 2020-01-09
DM-166 clean up user resource update

sutherlanda:
This PR also includes changes made in `feature/jesse/org-form`, and adds improvements to the user resource update for DM-166.

- Switch to ADT format
- Move inline validation (part of DM-177)
- Other improvements related to DM-168, DM-169, and DM-165

Leaving in draft for now as it will break front-end and avatar uploads until other tickets are ready.


dhruvio:
Hi Andrew, I won't get to reviewing this until Monday. I'll need to update the front-end code too — maybe we can pair or I can screen share if you're interested. Also, could you please merge development into your branch? The org forms have been merged into development.

sutherlanda:
Hey Dhruv, I've merged development into this branch, as well as my fixes for DM-189.

dhruvio:
It's going to take me a day or two to merge this because I need to update the front-end accordingly. I'll push my commits to this branch.

src/back-end/lib/db.ts
```diff
@@ -344,11 +343,14 @@ export async function readManyAffiliations(connection: Connection, userId: Id):
...
+    .select('affiliations.*')
     .where({
-      user,
-      organization
+      'affiliations.user': user,
+      'affiliations.organization': organization
     })
-    .andWhereNot({ membershipStatus: MembershipStatus.Inactive })
+    .andWhereNot({ 'affiliations.membershipStatus': MembershipStatus.Inactive })
+    .andWhereNot({ 'organizations.active': false })
```
dhruvio: I think you can combine these two `andWhereNot` calls into a single one by merging the argument objects.

src/back-end/lib/db.ts
```diff
@@ -357,8 +359,11 @@ export async function readOneAffiliation(connection: Connection, user: Id, organ
 
 export async function readOneAffiliationById(connection: Connection, id: Id): Promise<Affiliation> {
   const result = await connection('affiliations')
-    .where({ id })
-    .andWhereNot({ membershipStatus: MembershipStatus.Inactive })
+    .join('organizations', 'affiliations.organization', '=', 'organizations.id')
+    .select('affiliations.*')
+    .where({ 'affiliations.id': id })
+    .andWhereNot({ 'affiliations.membershipStatus': MembershipStatus.Inactive })
+    .andWhereNot({ 'organizations.active': false })
```
dhruvio: See my previous comment, `andWhereNot` usage can be merged

src/back-end/lib/db.ts
```diff
@@ -406,11 +417,20 @@ export async function approveAffiliation(connection: Connection, id: Id): Promis
...
+    .where({
+      id
+    })
+    .whereIn('organization', function() {
+      this.select('id')
+          .from('organizations')
+          .where({
+            active: true
+          });
+    });
```
dhruvio: Does the order of the `update` and `where*` calls matter?

src/back-end/lib/resources/affiliation.ts
```diff
@@ -58,63 +56,58 @@ const resource: Resource = {
 
   create(connection) {
     return {
-      async parseRequestBody(request) {
-        return request.body.tag === 'json' ? request.body.value : {};
+      async parseRequestBody(request): Promise<CreateRequestBody> {
```
dhruvio: You shouldn't need to type the return value of this function, as it should be inferred by the `Resource` type on line 31.

src/back-end/lib/resources/affiliation.ts
```diff
@@ -58,63 +56,58 @@ const resource: Resource = {
...
+        const body = request.body.tag === 'json' ? request.body.value : {};
+        return {
+          user: getString(body, 'user'),
+          organization: getString(body, 'organization'),
+          membershipType: getString(body, 'membershipType')
+        };
       },
-      async validateRequestBody(request) {
+      async validateRequestBody(request): Promise<Validation<ValidatedCreateRequestBody, CreateValidationErrors>> {
```
dhruvio: You shouldn't need to type the return value of this function, as it should be inferred by the `Resource` type on lines 30 & 31.

src/back-end/lib/resources/session.ts
```diff
@@ -43,7 +43,7 @@ export async function signOut(connection: Connection, session: Session): Promise
       ClientHttpMethod.Post,
       `${KEYCLOAK_URL}/auth/realms/${KEYCLOAK_REALM}/protocol/openid-connect/logout`,
       formData,
-      { 'Content-Type': 'application/x-www-form-urlencdoed'}
+      { 'Content-Type': 'application/x-www-form-urlencoded'}
```
dhruvio: Thanks!

src/back-end/lib/routers/auth.ts
```diff
@@ -134,7 +134,8 @@ async function makeRouter(connection: Connection): Promise<Router<any, any, any,
               type: userType,
               status: UserStatus.Active,
               name: claims.name || '',
-              email: claims.email,
+              email: claims.email || '',
+              jobTitle: '', // TODO - retrieve job title from claims for IDIR
```
dhruvio: Can remove the `TODO` comment

src/back-end/lib/resources/affiliation.ts
```diff
@@ -131,40 +124,38 @@ const resource: Resource = {
...
+        if (isInvalid(validatedAffiliation)) {
           return invalid({
-            affiliation: getInvalidValue(validatedAffiliationId, undefined)
+            affiliation: getInvalidValue(validatedAffiliation, undefined)
           });
         }
-        const existingAffiliation = validatedAffiliationId.value;
+        const existingAffiliation = validatedAffiliation.value;
         if (existingAffiliation.membershipStatus === MembershipStatus.Pending) {
           if (!permissions.updateAffiliation(connection, request.session, existingAffiliation.organization.id)) {
```
dhruvio: You need to `await permissions.updateAffiliation...` because it returns a promise.

src/back-end/lib/resources/affiliation.ts
```diff
@@ -58,63 +56,58 @@ const resource: Resource = {
...
               });
             }
-            return valid({ ...validResponseObject, membershipType, membershipStatus: MembershipStatus.Active });
+            // If existing, active affiliation, create new affiliation with ACTIVE status and updated role
+            return valid({
+              user,
+              organization,
+              membershipType: (validatedMembershipType.value as MembershipType),
+              membershipStatus: MembershipStatus.Active
+            });
```
dhruvio: I don't understand the use-case for this condition? Why would we simply create a second active affiliation for a user? Also, it doesn't look like the existing affiliation is checked to be active per the comment.

src/back-end/lib/resources/affiliation.ts
```diff
@@ -58,63 +56,58 @@ const resource: Resource = {
...
-            if (validatedMembershipType.tag === 'invalid') {
+            // If no existing, active affiliation, create new affiliation with PENDING status
+            return valid({
+              user,
+              organization,
+              membershipType: (validatedMembershipType.value as MembershipType),
+              membershipStatus: MembershipStatus.Pending
+            });
+          } else {
+            if (!permissions.updateAffiliation(connection, request.session, organization)) {
```
dhruvio: Need to `await` the permissions function call because it returns a promise.

src/back-end/lib/resources/affiliation.ts
dhruvio: I may have forgotten in case we discussed this previously :)

src/back-end/lib/resources/affiliation.ts
```diff
@@ -131,40 +124,38 @@ const resource: Resource = {
 
   update(connection) {
     return {
-      async parseRequestBody(request) {
+      async parseRequestBody(request): Promise<null> {
```
dhruvio: You shouldn't need to type the return value of this function, as it should be inferred by the `Resource` type above.

src/back-end/lib/resources/affiliation.ts
```diff
@@ -131,40 +124,38 @@ const resource: Resource = {
...
         return null;
       },
-      async validateRequestBody(request) {
-        const validatedAffiliationId = await validateAffiliationId(connection, request.params.id);
-        if (validatedAffiliationId.tag === 'invalid') {
+      async validateRequestBody(request): Promise<Validation<ValidatedUpdateRequestBody, UpdateValidationErrors>> {
```
dhruvio: You shouldn't need to type the return value of this function, as it should be inferred by the `Resource` type above.

src/back-end/lib/resources/affiliation.ts
```diff
@@ -131,40 +124,38 @@ const resource: Resource = {
...
+          return valid(existingAffiliation.id);
         }
         return invalid({
-          membershipType: ['Membership is not pending.']
+          affiliation: ['Membership is not pending.']
         });
       },
-      async respond(request) {
-        const respond = (code: number, body: Affiliation | UpdateValidationErrors | string[]) => basicResponse(code, request.session, makeJsonResponseBody(body));
+      async respond(request): Promise<Response<JsonResponseBody<Affiliation | UpdateValidationErrors>, Session>> {
```
dhruvio: This type signature is good because the `Resource` type only types the response body as one of `SupportedResponseBodies` which is very vague.

src/back-end/lib/resources/affiliation.ts
```diff
@@ -58,63 +56,58 @@ const resource: Resource = {
...
+          });
         }
-
-        return invalid({
-          membershipType: ['Invalid membership type specified']
-        });
       },
-      async respond(request) {
-        const respond = (code: number, body: Affiliation | string[]) => basicResponse(code, request.session, makeJsonResponseBody(body));
+      async respond(request): Promise<Response<JsonResponseBody<Affiliation | CreateValidationErrors>, Session>> {
```
dhruvio: This type signature is good because the `Resource` type only types the response body as one of `SupportedResponseBodies` which is very vague.

src/back-end/lib/resources/affiliation.ts
```diff
@@ -173,7 +164,7 @@ const resource: Resource = {
 
   delete(connection) {
     return {
-      async validateRequestBody(request) {
+      async validateRequestBody(request): Promise<Validation<ValidatedDeleteRequestBody, DeleteValidationErrors>> {
```
dhruvio: You shouldn't need to type the return value of this function, as it should be inferred by the Resource type above.

dhruvio: Also, there is no permissions check for deleting?

src/back-end/lib/resources/affiliation.ts
```diff
@@ -194,19 +186,17 @@ const resource: Resource = {
           });
```
dhruvio: You need to `await` the permissions call.

src/back-end/lib/resources/affiliation.ts
```diff
@@ -194,19 +186,17 @@ const resource: Resource = {
...
-        return valid({
-          affiliation: existingAffiliation.id
-        });
+        return valid(existingAffiliation.id);
       },
-      async respond(request) {
-        const respond = (code: number, body: Affiliation | DeleteValidationErrors | string[]) => basicResponse(code, request.session, makeJsonResponseBody(body));
+      async respond(request): Promise<Response<JsonResponseBody<Affiliation | DeleteValidationErrors>, Session>> {
+        const respond = (code: number, body: Affiliation | DeleteValidationErrors) => basicResponse(code, request.session, makeJsonResponseBody(body));
         if (request.body.tag === 'invalid') {
```
dhruvio: use `isInvalid` from `shared/lib/validations`

src/shared/lib/validation/user.ts
```diff
@@ -15,16 +15,16 @@ export function validateNotificationsOn(notificationsOn: boolean): Validation<Da
   return notificationsOn ? valid(new Date()) : valid(null);
 }
 
-export function validateAcceptedTerms(acceptedTerms: boolean): Validation<Date> {
-  return acceptedTerms ? valid(new Date()) : invalid(['You cannot unaccept the terms and conditions.']);
+export function validateAdminStatus(adminStatus: boolean): Validation<UserType> {
+  return adminStatus ? valid(UserType.Admin) : valid(UserType.Government);
```
dhruvio: I don't think this needs to be a validation since it takes `boolean` and only returns valid values. My suggestion is to simply create a helper function that converts a boolean to `UserType`. Something like `adminPermissionsToUserType(admin: boolean): UserType`

src/shared/lib/resources/user.ts
```diff
@@ -57,21 +57,29 @@ export function isPublicSectorEmployee(user: User): boolean {
...
+type UpdateProfileValidationErrors = ErrorTypeFrom<UpdateProfileRequestBody>;
+
+export type UpdateValidationErrors
+  = ADT<'updateProfile', UpdateProfileValidationErrors>
+  | ADT<'acceptTerms', string[]>
+  | ADT<'updateNotifications', string[]>
+  | ADT<'updateAdminPermissions', string[]>
+  | ADT<'parseFailure'>
+  | ADT<'permissions', string[]>
+  | ADT<'userNotFound', string[]>;
```
dhruvio: Types look great.

dhruvio:
Related ticket (DM-191) for necessary front-end changes: https://bcdevex.atlassian.net/jira/software/projects/DM/boards/4?assignee=5c591ea643b70c4fb64b4e55&selectedIssue=DM-191

Ticket has been assigned to me.

src/back-end/lib/db.ts
sutherlanda: It doesn't no, Knex will construct the SQL the same way regardless.

src/back-end/lib/resources/affiliation.ts
sutherlanda: So I thought that was the case, but some of the functions were not showing compile-time errors unless I explicitly put this in.  I'll take another look though, maybe I was not doing something correctly.

sutherlanda: Yeah, I confirmed that if I remove the typing of the return value, it will not type check the return value properly.  This was the source of the my previous errors that resulted in the need for DM-169 (Review CRUD resources to ensure the result from parseRequestBody matches the `Create/UpdateRequestBody` types in shared.).  Let me know if I am missing something, or if there is a better way to do this - seems like a compiler limitation with TypeScript.

src/back-end/lib/resources/affiliation.ts
sutherlanda: The idea was to create a new affiliation to replace their active one (if it exists), otherwise to create a new pending one if it doesn't.  I agree something isn't quite right here though, and I'll take another look.

src/shared/lib/validation/user.ts
sutherlanda: Yes, I will do this.

sutherlanda:
> I have reviewed all files except for the back-end user resource (will look at it tomorrow). There some questions and changes I've requested via my comments.
> 
> Another change that I'd recommend is in the back-end permissions module. Some of the async permissions functions use boolean logic to test check for certain conditions without taking advantage of short-circuiting. For example, in an OR expression, the first check is an expensive check that performs some kind of I/O, and if that fails, subsequent checks are made that are much less expensive without I/O. The order of these operations should be flipped. i.e.:
> 
> `return await isUserOwnerOfOrg(connection, session.user, orgId) || isAdmin(session);`
> 
> Should become:
> 
> `return isAdmin(session) || await isUserOwnerOfOrg(connection, session.user, orgId);`
> 
> So, if the user is an admin, the `isUserOwnerOfOrg` function is never called. The first snippet results in the (more expensive) async operation to be called every time, even if the user is an admin.

Thanks, I will make this change as well.  I've replied to your first comment regarding explicitly defining function return types for crud actions.  I've fixed the straightforward ones, and will update the remaining ones tomorrow.

src/shared/lib/validation/user.ts
sutherlanda: Done, also switched validation of notifications flag to helper.

src/back-end/lib/resources/affiliation.ts
dhruvio: I think the reason the types were still passing were because `JsonRequestBody`'s values are types as `any` by default.

dhruvio: I'll try changing that to `unknown` to see if that solves the problem.

dhruvio: I ended up changing the `JsonRequestBody` type to use a new `JsonValue` type I defined. I couldn't do this until TypeScript 3.7, because the language didn't support recursive type aliases. Have a look at https://github.com/bcgov/digital_marketplace/pull/32/commits/6d94653948176d5e6ff0db3fb6e72bccb24504c9 if you are interested.

I have also removed the unnecessary `parseRequestBody` type signatures.

src/back-end/lib/resources/affiliation.ts
dhruvio: Let me know how you go.

src/back-end/lib/resources/affiliation.ts
sutherlanda: Ah, nice thanks for that.  I can clean up the rest of the unnecessary return types.

dhruvio:
Changes look pretty sweet. I have replied to some comments above. I still need to make the necessary front-end changes.

src/back-end/lib/resources/affiliation.ts
sutherlanda: Simplified it a bit - and added an extra check for pending state.  Still not ideal, I'm wondering if we should use an ADT to separate out the scenarios where want to approve a pending membership, and creating a new membership.  Let me know if I should do this - don't want to create more front-end work unless necessary.

dhruvio:
Great job. Thanks @sutherlanda!

PR #33, opened 2020-01-11
DM-163 implement file permissions validation in shared

sutherlanda:
This PR refactors the file permissions parsing and adds a file permissions validation function in shared.

- Also cleaned up file resource to align with requirements of DM-168, DM-169.

** This branch is off of `feature/andrew/parse-type-match`.

### DIFF
--- a/src/back-end/lib/db.ts
+++ b/src/back-end/lib/db.ts
@@ -13,11 +13,11 @@ import { adt, Id } from 'shared/lib/types';
 import { invalid, isValid, valid } from 'shared/lib/validation';
 import { DatabaseValidation } from 'shared/lib/validation/db';
 
-export type Connection = Knex<any, any>;
+export type Connection = Knex;
 
 const createDatabaseError = () => adt('databaseError' as const);
 
-export async function createUser(connection: Connection, user: Omit<User, 'id' | 'notificationsOn' | 'acceptedTerms'>): Promise<User> {
+export async function createUser(connection: Connection, user: Omit<User, 'id' | 'notificationsOn' | 'acceptedTerms' | 'deactivatedOn' | 'deactivatedBy'>): Promise<User> {
   const now = new Date();
   const [result] = await connection('users')
     .insert({
@@ -32,15 +32,17 @@ export async function createUser(connection: Connection, user: Omit<User, 'id' |
   return await rawUserToUser(connection, result);
 }
 
-export async function updateUser(connection: Connection, userInfo: Omit<Partial<User>, 'avatarImageFile'> & { id: Id, avatarImageFile?: Id }): Promise<DatabaseValidation<User>> {
+type UpdateUserParams = Omit<Partial<User>, 'avatarImageFile'> & { id: Id, avatarImageFile?: Id };
+
+export async function updateUser(connection: Connection, userInfo: UpdateUserParams): Promise<DatabaseValidation<User>> {
   const now = new Date();
   try {
-    const result: RawUserToUserParams = await connection('users')
+    const [result] = await connection<RawUserToUserParams>('users')
     .where({ id: userInfo.id })
     .update({
       ...userInfo,
       updatedAt: now
-    }, ['*']);
+    } as UpdateUserParams, '*');
     if (!result) {
       throw new Error('unable to update user');
     }
@@ -52,9 +54,9 @@ export async function updateUser(connection: Connection, userInfo: Omit<Partial<
 
 export async function readOneUser(connection: Connection, id: Id): Promise<DatabaseValidation<User | null>> {
   try {
-    const result: RawUserToUserParams = await connection('users')
-    .where({ id })
-    .first();
+    const result = await connection<RawUserToUserParams>('users')
+      .where({ id })
+      .first();
     return valid(result ? await rawUserToUser(connection, result) : null);
   } catch (exception) {
     return invalid(createDatabaseError());
@@ -63,7 +65,7 @@ export async function readOneUser(connection: Connection, id: Id): Promise<Datab
 
 export async function readManyUsers(connection: Connection): Promise<DatabaseValidation<User[]>> {
   try {
-    const results: RawUserToUserParams[] = await connection('users').select();
+    const results = await connection<RawUserToUserParams>('users').select();
     return valid(await Promise.all(results.map(async raw => await rawUserToUser(connection, raw))));
   } catch (exception) {
     return invalid(createDatabaseError());
@@ -71,12 +73,12 @@ export async function readManyUsers(connection: Connection): Promise<DatabaseVal
 }
 
 interface RawUserToUserParams extends Omit<User, 'avatarImageFile'> {
-  avatarImageFile: Id;
+  avatarImageFile: Id | null;
 }
 
 export async function rawUserToUser(connection: Connection, params: RawUserToUserParams): Promise<User> {
   const { avatarImageFile: fileId, ...restOfRawUser } = params;
-  const avatarImageFile = fileId && await readOneFileById(connection, fileId) || undefined;
+  const avatarImageFile = fileId && await readOneFileById(connection, fileId) || null;
   return {
     ...restOfRawUser,
     avatarImageFile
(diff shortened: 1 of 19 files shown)


src/back-end/lib/resources/avatar.ts
```diff
@@ -0,0 +1,72 @@
...
+  null,
+  null,
+  Session,
+  Connection
+>;
+
+const resource: Resource = {
+  routeNamespace: 'avatars',
+
+  readOne: FileResource.readOne,
```
dhruvio: Let's remove this and simply use `/api/files/:id` to read one file. This resource can just handle creating an avatar.

src/shared/lib/resources/file.ts
```diff
@@ -38,46 +39,20 @@ export type FilePermissions<Id, UserType>
...
-    default:
-      return null;
+  // Validate each file permission
+  const validatedFilePermissions = validateArray(raw, validateFilePermission);
+  if (isValid(validatedFilePermissions)) {
+    // Reduce array of valid permissions into a unique set
+    const uniquePerms = Array.from(new Set(validatedFilePermissions.value.map(v => JSON.stringify(v))))
+      .map(v => JSON.parse(v) as FilePermissions<Id, UserType>);
+    return uniquePerms;
   }
```
dhruvio: This should be moved into a separate function titled `validateFilePermissions` in `shared/lib/validation/file`. That function should include deduping the array of permissions. Then, `parseFilePermissions` can be a lot simpler.

dhruvio: If we have time, at a later date, we could update the server adapter to support validating file metadata instead of just parsing them so API consumers are aware of why the file metadata is invalid.

src/shared/lib/resources/user.ts
```diff
@@ -37,12 +38,12 @@ export interface User {
...
-  notificationsOn?: Date;
-  acceptedTerms?: Date;
+  avatarImageFile: FileRecord | null;
+  notificationsOn: Date | null;
+  acceptedTerms: Date | null;
   idpUsername: string;
-  deactivatedOn?: Date;
-  deactivatedBy?: Id;
+  deactivatedOn: Date | null;
+  deactivatedBy: Id | null;
```
dhruvio: @sutherlanda TODO: investigate whether this change needs to be made across the board.

src/shared/lib/validation/db.ts
```diff
@@ -0,0 +1,8 @@
+import { ADT } from 'shared/lib/types';
+import { Validation } from 'shared/lib/validation';
+
+export type DatabaseError
+  = ADT<'databaseError'>;
+
+export type DatabaseValidation<Valid, Invalid = DatabaseError>
+  = Validation<Valid, Invalid>;
```
dhruvio: We talked about some potential changes here:

```ts
export type DatabaseValidation<Valid> = Validation<Valid, null>;
// invalid simply corresponds to an ambiguous database error.
```

src/shared/lib/validation/file.ts
```diff
@@ -1,5 +1,44 @@
...
+import { adt, Id } from 'shared/lib/types';
+import { invalid, isValid, mapValid, valid, validateGenericString, validateUUID, Validation } from 'shared/lib/validation';
+import { isString } from 'util';
 
-export function validateFileName(name: string): Validation<string> {
-  return validateGenericString(name, 'File name');
+export function validateFileName(name: string, validExtensions: string[] = []): Validation<string> {
+  const validatedName = validateGenericString(name, 'File name');
+  if (isValid(validatedName) && validExtensions.length > 0) {
+    const extension = name.substr(name.lastIndexOf('.') + 1);
```
dhruvio: Change `name` to `validatedName.value`. Also, what if the name doesn't contain a `'.'`?

Might be safer to use a regexp:

```ts
//...
const extension = name.match(/\.([a-zA-Z0-9]+)$/);
if (extension && validExtensions.map(ext => ext.toLowerCase()).includes(extension[1].toLowerCase())) {
//...
```

src/shared/lib/validation/file.ts
```diff
@@ -1,5 +1,44 @@
...
+      return invalid(['Invalid file extension.']);
+    }
+  } else {
+    return validatedName;
+  }
+}
+
+export function validateFilePath(path: string): Validation<string> {
+  return validateGenericString(path, 'File path');
+}
```
dhruvio: This can be moved to `back-end/lib/validation` since the front-end doesn't use it.

src/shared/lib/validation/file.ts
```diff
@@ -1,5 +1,44 @@
...
+    case 'userType':
+      const userType = parseUserType(raw.value);
+      if (userType) {
+        return valid(adt('userType', userType));
+      } else {
+        return invalid(['Not a valid user type.']);
+      }
+    default:
+      return invalid(['Invalid file permission provided.']);
+  }
```
dhruvio: 👍

src/back-end/lib/db.ts
```diff
@@ -28,64 +32,72 @@ export async function createUser(connection: Connection, user: Omit<User, 'id' |
...
+export async function readOneUser(connection: Connection, id: Id): Promise<DatabaseValidation<User | null>> {
+  try {
+    const result = await connection<RawUserToUserParams>('users')
+      .where({ id })
+      .first();
+    return valid(result ? await rawUserToUser(connection, result) : null);
+  } catch (exception) {
+    return invalid(createDatabaseError());
+  }
 }
```
dhruvio: You may want to create a helper function to standardize this error handling behaviour:

```ts
type DbValidation<Valid> = Validation<Valid, null>;

type DbFn<Arg, Valid> = (connection: Connection, arg: Arg) => Promise<DatabaseValidation<Valid>>;

function tryDb<Arg, Valid>(fn: DbFn<Arg, Valid>): DbFn<Arg, Valid> { 
  return async (connection, arg) => { 
    try { 
      return await fn(connection, arg);
    } catch (e) { 
      return invalid(null);
    } 
  };
}

// Usage
export const readOneUser = tryDb<Id /*Arg*/, User /*Valid*/>(userId /*Arg*/ => async connection => { 
  const result = await connection<RawUserToUserParams>('users')
    .where({ id })
    .first();
  return valid(result ? await rawUserToUser(connection, result) : null);
});
```

src/back-end/lib/permissions.ts
```diff
@@ -139,3 +139,9 @@ export function createFile(session: Session): boolean {
 export async function readOneFile(connection: Connection, session: Session, fileId: string): Promise<boolean> {
   return await hasFilePermission(connection, session, fileId);
 }
+
+// Avatars.
+
+export function createAvatar(session: Session, userId: string): boolean {
+  return isUser(session) && isOwnAccount(session, userId);
+}
```
dhruvio: I think this can be removed, see previous related comment

dhruvio:
Note to self: TODO Update front-end code to use new `avatars#create` CRUD action.

src/shared/lib/resources/file.ts
sutherlanda: Sounds good - I added in proper invalid string values in the return from `validateFilePermission` in anticipation of this.

src/shared/lib/validation/file.ts
```diff
@@ -19,8 +23,14 @@ export function validateFileName(name: string, validExtensions: string[] = []):
...
-  return validateGenericString(path, 'File path');
+export function validateFilePermissions(raw: any): ArrayValidation<FilePermissions<Id, UserType>> {
+  raw = isArray(raw) ? raw : [raw];
+  const validatedFilePermissions = validateArray(raw, validateFilePermission);
+  return mapValid(validatedFilePermissions, perms => {
+    return Array
+      .from(new Set(perms.map(v => JSON.stringify(v))))
+      .map(v => JSON.parse(v) as FilePermissions<Id, UserType>);
+  });
 }
```
dhruvio: @sutherlanda I refactored this function to take advantage of validation's helper functions (i.e. mapValid). The semantic change was that non-array raw values are coerced into a single-element array.

PR #37, opened 2020-01-17
DM-186 Support redirectOnSuccess for sign in

jjbandit:


src/front-end/typescript/lib/app/router.ts
```diff
@@ -111,10 +111,10 @@ const router: Router<Route> = {
     },
     {
       path: '/sign-in',
-      makeRoute() {
+      makeRoute({query}) {
         return {
           tag: 'signIn',
-          value: null
+          value: { redirectOnSuccess: query.redirectOnSuccess as string } // TODO(Jesse): Why does this need to be cast?
```
dhruvio: In `framework.tsx`, the values of the query object can be `string | string[] | undefined`. You should use `getString(query, 'redirectOnSuccess') || ''` from `shared/lib`.

src/front-end/typescript/lib/app/router.ts
```diff
@@ -175,7 +175,7 @@ const router: Router<Route> = {
       case 'content':
         return `/content/${route.value}`;
       case 'signIn':
-        return '/sign-in';
+        return `/sign-in${route.value.redirectOnSuccess ? `?redirectOnSuccess=${window.encodeURIComponent(route.value.redirectOnSuccess)}` : ''}`;
```
dhruvio: 👍

PR #38, opened 2020-01-18
DM-177 change all db routines to return validation

sutherlanda:
This PR updates the db layer routines to return Validation.  All callers of database functions have been updated to check validity of responses and handle them accordingly.

CRUD handlers will return a 503 (database not available) if a database exception occurs.  This shouldn't occur under normal circumstances, but if it does, a stack trace will not be returned to the client.

@dhruvio this ended up being a larger change than I anticipated.  I started in branch `feature/andrew/file-resource-cleanup` but put the majority of the changes in this PR for an easier review.  Let me know if you'd like me to just put everything in one PR.


dhruvio:
@sutherlanda Please remove the "WIP" label when this is ready for review.

src/back-end/lib/crud.ts
```diff
@@ -86,3 +87,37 @@ export function makeRouter<SupportedRequestBodies, SupportedResponseBodies, Crea
...
+  permissions?: string[];
+  database?: string[];
+}
+
+interface ResponseValidation<ValidBody, InvalidBody, ResB, Session> {
+  valid: Respond<ValidBody, ResB, Session>;
+  invalid: Respond<InvalidBody, ResB, Session>;
+}
+
+export function wrapRespond<ValidBody, InvalidBody extends BodyWithErrors, ResB, Session>(responseValidation: ResponseValidation<ValidBody, InvalidBody, ResB, Session>): Respond<Validation<ValidBody, InvalidBody>, ResB, Session> {
```
sutherlanda: @dhruvio can you confirm this is correct? Had a bit of trouble getting the types working for the inner `respond` function. Wouldn't mind doing a quick review with you before using through the CRUD resources.

src/back-end/lib/crud.ts
```diff
@@ -86,3 +87,37 @@ export function makeRouter<SupportedRequestBodies, SupportedResponseBodies, Crea
     return routes;
   };
 }
+
+interface BodyWithErrors {
+  permissions?: string[];
+  database?: string[];
+}
```
dhruvio: I suggest moving this to shared/lib/index.ts so you can extend this interface in the shared validation error interfaces

src/back-end/lib/crud.ts
dhruvio: Had a quick look, and it looks great!

sutherlanda:
Ready for @dhruvio.  Thanks!

src/back-end/lib/resources/affiliation.ts
```diff
@@ -103,19 +109,18 @@ const resource: Resource = {
...
-            if (request.body.value.permissions) {
-              return basicResponse(401, request.session, makeJsonResponseBody(request.body.value));
-            }
-            return basicResponse(400, request.session, makeJsonResponseBody(request.body.value));
-          case 'valid':
-            const affiliation = await createAffiliation(connection, request.body.value);
-            return respond(201, affiliation);
-        }
-      }
+      respond: wrapRespond<ValidatedCreateRequestBody, CreateValidationErrors, JsonResponseBody<Affiliation>, JsonResponseBody<CreateValidationErrors>, Session>({
```
sutherlanda: I had to specify the type parameters here to prevent the error I sent you earlier.  I still don't completely understand why, but I believe it was having trouble inferring a `ValidResB` from a `InvalidResB`.  Would be interested to know if you had any insight on why this was.  This only occurred on the create actions for `affiliation` and other resources.

sutherlanda:
Ready for review again @dhruvio.  Left a comment outlining the type issue we were discussing earlier, and the workaround.  Thanks.

src/back-end/lib/resources/affiliation.ts
dhruvio: This is fine. TypeScript's type inference isn't good enough to figure out what the return type should be.

PR #40, opened 2020-01-26
DM 211 database migration for CWU

sutherlanda:
This PR includes a database migration for the CWU opportunity, proposal, and related data models.

The designs for this model can be viewed in the shared Google Drive.

*Note: This is based on branch `feature/andrew/refactor-affiliation` because I needed a migration that was included in that branch.  Recommend merging branches `feature/andrew/db-validate` and `feature/andrew/refactor-affiliation` into `development` prior to reviewing this PR.

### DIFF
--- a/src/migrations/tasks/20200126113825_cwu_init.ts
+++ b/src/migrations/tasks/20200126113825_cwu_init.ts
@@ -1,0 +1,163 @@
+import { makeDomainLogger } from 'back-end/lib/logger';
+import { console as consoleAdapter } from 'back-end/lib/logger/adapters';
+import Knex from 'knex';
+import { OpportunityStatus } from 'shared/lib/resources/cwu-opportunity';
+import { ProposalStatus } from 'shared/lib/resources/cwu-proposal';
+
+const logger = makeDomainLogger(consoleAdapter, 'migrations');
+
+export async function up(connection: Knex): Promise<void> {
+  // Create CWUOpportunity table
+  await connection.schema.createTable('cwuOpportunities', table => {
+    table.uuid('id').primary().unique().notNullable();
+    table.timestamp('createdAt').notNullable();
+    table.timestamp('updatedAt').notNullable();
+    table.uuid('createdBy').references('id').inTable('users');
+    table.uuid('updatedBy').references('id').inTable('users');
+  });
+  logger.info('Created cwuOpportunities table.');
+
+  // Create CWUOpportunityVersion table
+  await connection.schema.createTable('cwuOpportunityVersions', table => {
+    table.uuid('id').primary().unique().notNullable();
+    table.timestamp('createdAt').notNullable();
+    table.uuid('createdBy').references('id').inTable('users').notNullable();
+    table.uuid('opportunity').references('id').inTable('cwuOpportunities');
+    table.string('title');
+    table.string('teaser');
+    table.boolean('remoteOk').defaultTo(false);
+    table.string('remoteDesc');
+    table.string('location');
+    table.integer('reward');
+    table.specificType('skills', 'text ARRAY');
+    table.string('description');
+    table.timestamp('proposalDeadline');
+    table.timestamp('assignmentDate');
+    table.timestamp('startDate');
+    table.timestamp('completionDate');
+    table.string('submissionInfo');
+    table.string('acceptanceCriteria');
+    table.string('evaluationCriteria');
+  });
+  logger.info('Created cwuOpportunityVersions table.');
+
+  // Create CWUOpportunityStatus table
+  await connection.schema.createTable('cwuOpportunityStatuses', table => {
+    table.uuid('id').primary().unique().notNullable();
+    table.timestamp('createdAt').notNullable();
+    table.uuid('createdBy').references('id').inTable('users').notNullable();
+    table.uuid('opportunity').references('id').inTable('cwuOpportunities');
+    table.enu('status', Object.values(OpportunityStatus)).notNullable();
+    table.string('note');
+  });
+  logger.info('Created cwuOpportunityStatuses table.');
+
+  // Create CWUOpportunityAttachment table
+  await connection.schema.createTable('cwuOpportunityAttachments', table => {
+    table.uuid('id').primary().unique().notNullable();
+    table.timestamp('createdAt').notNullable();
+    table.uuid('createdBy').references('id').inTable('users').notNullable();
+    table.uuid('opportunityVersion').references('id').inTable('cwuOpportunityVersions').notNullable();
+    table.uuid('file').references('id').inTable('files').notNullable();
+  });
+  logger.info('Created cwuOpportunityAttachments table.');
+
+  // Create CWUOpportunityAddendum table
+  await connection.schema.createTable('cwuOpportunityAddenda', table => {
+    table.uuid('id').primary().unique().notNullable();
+    table.uuid('opportunity').references('id').inTable('cwuOpportunities').notNullable();
+    table.timestamp('createdAt').notNullable();
+    table.uuid('createdBy').references('id').inTable('users').notNullable();
+    table.string('description').notNullable();
+  });
+  logger.info('Created cwuOpportunityAddenda table.');
+
+  // Create CWUProposal table
+  await connection.schema.createTable('cwuProposals', table => {
+    table.uuid('id').primary().unique().notNullable();
+    table.timestamp('createdAt').notNullable();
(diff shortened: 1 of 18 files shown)


src/migrations/tasks/20200126113825_cwu_init.ts
```diff
@@ -0,0 +1,163 @@
...
+const logger = makeDomainLogger(consoleAdapter, 'migrations');
+
+export async function up(connection: Knex): Promise<void> {
+  // Create CWUOpportunity table
+  await connection.schema.createTable('cwuOpportunities', table => {
+    table.uuid('id').primary().unique().notNullable();
+    table.timestamp('createdAt').notNullable();
+    table.timestamp('updatedAt').notNullable();
+    table.uuid('createdBy').references('id').inTable('users');
+    table.uuid('updatedBy').references('id').inTable('users');
```
dhruvio: We don't need the `updatedAt` and `updatedBy` columns

dhruvio: They are implicitly provided by opportunity versions

dhruvio: `createdBy` should be `notNullable`

src/migrations/tasks/20200126113825_cwu_init.ts
```diff
@@ -0,0 +1,163 @@
...
+    table.integer('reward');
+    table.specificType('skills', 'text ARRAY');
+    table.string('description');
+    table.timestamp('proposalDeadline');
+    table.timestamp('assignmentDate');
+    table.timestamp('startDate');
+    table.timestamp('completionDate');
+    table.string('submissionInfo');
+    table.string('acceptanceCriteria');
+    table.string('evaluationCriteria');
```
dhruvio: All of these columns, except `completionDate`, should be `notNullable`

src/migrations/tasks/20200126113825_cwu_init.ts
```diff
@@ -0,0 +1,163 @@
...
+  });
+  logger.info('Created cwuOpportunityVersions table.');
+
+  // Create CWUOpportunityStatus table
+  await connection.schema.createTable('cwuOpportunityStatuses', table => {
+    table.uuid('id').primary().unique().notNullable();
+    table.timestamp('createdAt').notNullable();
+    table.uuid('createdBy').references('id').inTable('users').notNullable();
+    table.uuid('opportunity').references('id').inTable('cwuOpportunities');
```
dhruvio: `notNullable`

src/migrations/tasks/20200126113825_cwu_init.ts
```diff
@@ -0,0 +1,163 @@
...
+    table.enu('status', Object.values(CWUOpportunityStatus)).notNullable();
+    table.string('note');
```
dhruvio: `notNullable`

src/migrations/tasks/20200126113825_cwu_init.ts
```diff
@@ -0,0 +1,163 @@
...
+  });
+  logger.info('Created cwuOpportunityStatuses table.');
+
+  // Create CWUOpportunityAttachment table
+  await connection.schema.createTable('cwuOpportunityAttachments', table => {
+    table.uuid('id').primary().unique().notNullable();
+    table.timestamp('createdAt').notNullable();
+    table.uuid('createdBy').references('id').inTable('users').notNullable();
```
dhruvio: We actually don't need these, as the `files` table provides `createdAt` and `createdBy`

src/migrations/tasks/20200126113825_cwu_init.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  // Create CWUProposal table
+  await connection.schema.createTable('cwuProposals', table => {
+    table.uuid('id').primary().unique().notNullable();
+    table.timestamp('createdAt').notNullable();
+    table.uuid('createdBy').references('id').inTable('users').notNullable();
+    table.timestamp('updatedAt').notNullable();
+    table.uuid('updatedBy').references('id').inTable('users').notNullable();
+    table.string('proposalText');
+    table.string('additionalComments');
```
dhruvio: `notNullable`

src/migrations/tasks/20200126113825_cwu_init.ts
```diff
@@ -0,0 +1,163 @@
...
+    table.string('mailCode').notNullable();
+    table.string('country').notNullable();
+  });
+  logger.info('Created cwuProponents table.');
+
+  // Add foreign key relationship Proposals --> Proponents
+  await connection.schema.alterTable('cwuProposals', table => {
+    table.uuid('proponent').references('id').inTable('cwuProponents');
+  });
+  logger.info('Foreign key \'proponent\' added to cwuProposals table.');
```
dhruvio: Why not create the `cwuProponents` table before the `cwuProposals` table so you don't need to alter it (you can add the `proponent` column to the `cwuProposals` table the first time)?

src/migrations/tasks/20200126113825_cwu_init.ts
```diff
@@ -0,0 +1,163 @@
...
+  // Add constraint to CWUProposal table (proponent is either individual or org - not both)
+  await connection.schema.raw(`
+  ALTER TABLE "cwuProposals"
+  ADD CONSTRAINT "eitherProponentOrOrg" check(
+    ("proponent" IS NOT NULL AND "proponentOrg" IS NULL)
+    OR
+    ("proponent" IS NULL AND "proponentOrg" IS NOT NULL)
+  )
+  `);
+  logger.info('Added constraint \'eitherProponentOrOrg\' to cwuProposals table.');
```
dhruvio: 👍 
Not super picky about this, but you could rename the `proponent` column to something like `proponentIndividual` for added clarity

src/migrations/tasks/20200126113825_cwu_init.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  // Create CWUProposalAttachment table
+  await connection.schema.createTable('cwuProposalAttachments', table => {
+    table.uuid('id').primary().unique().notNullable();
+    table.timestamp('createdAt').notNullable();
+    table.uuid('createdBy').references('id').inTable('users').notNullable();
```
dhruvio: `files` table provides these

src/migrations/tasks/20200126113825_cwu_init.ts
```diff
@@ -0,0 +1,163 @@
...
+  await connection.schema.dropTable('cwuProposalAttachments');
+  await connection.schema.raw(`
+    ALTER TABLE "cwuProposals"
+    DROP CONSTRAINT "eitherProponentOrOrg"
+  `);
+  await connection.schema.alterTable('cwuProposals', table => {
+    table.dropColumn('proponent');
+  });
+  await connection.schema.dropTable('cwuProponents');
+  await connection.schema.dropTable('cwuProposals');
```
dhruvio: I *think* you can drop the proposals table first and the proponents table second to avoid needing to drop the `proponent` column from the proposals table.

src/migrations/tasks/20200126113825_cwu_init.ts
sutherlanda: Didn't we decide that these would be nullable so that partial opportunities could be saved in DRAFT?  (Instead of having a separate schema for draft opportunities)

src/migrations/tasks/20200126113825_cwu_init.ts
sutherlanda: Yes, you're right.  I've also removed the `drop constraint` as that's taken care of by dropping the cwuProposals table.

src/migrations/tasks/20200126113825_cwu_init.ts
dhruvio: I think that was our original solution, but I thought we decided that the front-end would figure out how to send "minimal" values for drafts to avoid making the schema nullable?

sutherlanda: You are correct sir.  Will make the changes.

src/migrations/tasks/20200126113825_cwu_init.ts
sutherlanda: Good idea, done.

sutherlanda:
Thanks @dhruvio , changes completed.  Will also update data model diagram to match.

PR #44, opened 2020-01-31
DM-257 - back-end crud resource for CWU opportunities

sutherlanda:
This PR includes resource CRUD actions for CWU opportunities.

### DIFF
--- a/src/back-end/lib/db.ts
+++ b/src/back-end/lib/db.ts
@@ -712,32 +712,32 @@ async function rawCWUOpportunityToCWUOpportunity(connection: Connection, raw: Ra
 export const readManyCWUOpportunities = tryDb<[Session], CWUOpportunitySlim[]>(async (connection, session) => {
   // Retrieve the opportunity and most recent opportunity status
 
-  const results = await connection<RawCWUOpportunitySlim>('cwuOpportunities')
+  const results = await connection<RawCWUOpportunitySlim>('cwuOpportunities as opp')
     // Join on latest CWU status
-    .join<RawCWUOpportunitySlim>('cwuOpportunityStatuses', function() {
+    .join<RawCWUOpportunitySlim>('cwuOpportunityStatuses as stat', function() {
       this
-        .on('cwuOpportunities.id', '=', 'cwuOpportunityStatuses.opportunity')
-        .andOn('cwuOpportunityStatuses.createdAt', '=',
-          connection.raw('(select max(createdAt) from cwuOpportunityStatuses where \
-            cwuOpportunities.id = cwuOpportunityStatuses.opportunity)'));
+        .on('opp.id', '=', 'stat.opportunity')
+        .andOn('stat.createdAt', '=',
+          connection.raw('(select max("createdAt") from "cwuOpportunityStatuses" as stat2 where \
+            stat2.opportunity = opp.id)'));
     })
     // Join on latest CWU version
-    .join<RawCWUOpportunitySlim>('cwuOpportunityVersions', function() {
+    .join<RawCWUOpportunitySlim>('cwuOpportunityVersions as version', function() {
       this
-        .on('cwuOpportunities.id', '=', 'cuwOpportunityVersions.opportunity')
-        .andOn('cwuOpportunityVersions.createdAt', '=',
-          connection.raw('(select max(createdAt) from cwuOpportunityVersions where \
-            cwuOpportunities.id = cwuOpportunityVersions.opportunity)'));
+        .on('opp.id', '=', 'version.opportunity')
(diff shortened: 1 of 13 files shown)


sutherlanda:
@dhruvio this is ready for review

src/back-end/lib/db.ts
```diff
@@ -654,3 +656,434 @@ export async function hasFilePermission(connection: Connection, session: Session
...
+
+  const results = await query;
+  // Remove createdBy/updatedBy for non-admin or non-author
+  results.map(result => {
+    if (session?.user?.type !== UserType.Admin &&
+      session?.user?.id !== result.createdBy &&
+      session?.user?.id !== result.updatedBy) {
+        delete result.createdBy;
+        delete result.updatedBy;
+    }
```
dhruvio: If the user isn't signed in, the `createdBy` and `updatedBy` properties will not be deleted. They should be deleted for anonymous users.

src/back-end/lib/db.ts
```diff
@@ -654,3 +656,434 @@ export async function hasFilePermission(connection: Connection, session: Session
...
+      .where({ opportunity: id })
+      .select('id')).map(row => row.id);
+
+    // Remove createdBy/updatedBy for non-admin or non-author
+    if (session?.user?.type !== UserType.Admin &&
+      session?.user?.id !== result.createdBy &&
+      session?.user?.id !== result.updatedBy) {
+        delete result.createdBy;
+        delete result.updatedBy;
+    }
```
dhruvio: Can we make this code more DRY?

dhruvio: Maybe move it to `rawCWUOpportunitytoCWUOpportunity`

src/back-end/lib/db.ts
```diff
@@ -654,3 +656,434 @@ export async function hasFilePermission(connection: Connection, session: Session
...
+}
+
+export const updateCWUOpportunityStatus = tryDb<[Id, CWUOpportunityStatus, string, Session], CWUOpportunity>(async (connection, id, status, note, session) => {
+  const now = new Date();
+  const [result] = await connection<CWUOpportunityStatusRecord>('cwuOpportunityStatuses')
+    .insert({
+      id: generateUuid(),
+      opportunity: id,
+      createdAt: now,
+      createdBy: session.user?.id,
```
dhruvio: This entire operation should fail if the user is not signed in — shouldn't use optional chaining here.

src/back-end/lib/db.ts
```diff
@@ -654,3 +656,434 @@ export async function hasFilePermission(connection: Connection, session: Session
...
+}
+
+export const addCWUOpportunityAddendum = tryDb<[Id, string, Session], CWUOpportunity>(async (connection, id, addendumText, session) => {
+  const now = new Date();
+  const [result] = await connection<CWUOpportunityAddendumRecord>('cwuOpportunityAddenda')
+    .insert({
+      id: generateUuid(),
+      opportunity: id,
+      description: addendumText,
+      createdBy: session.user?.id,
```
dhruvio: See note above about optional chaining — this should fail if the user is not signed in.

src/shared/lib/resources/code-with-us.ts
```diff
@@ -38,4 +39,70 @@ export interface CWUOpportunity {
...
+    case CWUOpportunityStatus.Suspended:
+      return [CWUOpportunityStatus.Published, CWUOpportunityStatus.Canceled].includes(to);
+    default:
+      return false;
+  }
+}
+
+export const publicOpportunityStatuses = [CWUOpportunityStatus.Published, CWUOpportunityStatus.Evaluation, CWUOpportunityStatus.Awarded];
+
+export const privateOpportunitiesStatuses = [CWUOpportunityStatus.Draft, CWUOpportunityStatus.Canceled, CWUOpportunityStatus.Suspended];
```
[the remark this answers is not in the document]
sutherlanda: I made these readonly types, but Knex complained on it's `whereIn` calls as it doesn't guarantee readonly on it's parameters (even though it should).  I've casted on those calls to compensate, i.e.

`whereIn('stat.status', publicOpportunityStatuses as CWUOpportunityStatus[])`

src/back-end/lib/db.ts
sutherlanda: I extracted this out into a function.  Couldn't move directly into `rawCWUOpportunitytoCWUOpportunity` since it's also needed in `rawCWUOpportunitySlimtoCWUOpportunitySlim` but I think this is still an improvement.  Let me know if you'd prefer a different approach.

sutherlanda:
Thanks for the review @dhruvio !.  Implemented the changes requested 👍

PR #47, opened 2020-02-14
DM-266 update opportunity status

sutherlanda:
* Allow nullable on createdBy


dhruvio:
Can you also implement a `UserSlim` type that contains the user' ID and name only? We can use that for `createdBy` fields. For CWU, government opportunity owners shouldn't be able to view an entire user's profile via this field, just their name and ID.

sutherlanda:
> Can you also implement a `UserSlim` type that contains the user' ID and name only? We can use that for `createdBy` fields. For CWU, government opportunity owners shouldn't be able to view an entire user's profile via this field, just their name and ID.

Done. I've used `UserSlim` for all createdBy and updatedBy fields.  There was one front-end error in `history.tsx` that I fixed, but you may want to double-check that.

PR #48, opened 2020-02-17
DM-227 add opportunity subscriber resource

sutherlanda:
This PR adds a resource for opportunity subscribers/watchers.  It includes endpoints for deleting and creating subscriptions.

The create request body consists of an opportunity id, and will create a subscription for the current user (pulled from session)

The delete route should includes a URL param for the opportunity id (since delete does not have a request body).

Also updated CWUOpportunity type to include an optional `subscribed` property (only defined for authenticated users).

### DIFF
--- a/src/back-end/lib/server/index.ts
+++ b/src/back-end/lib/server/index.ts
@@ -423,6 +423,8 @@ export function wrapRespond<ValidReqB, InvalidReqB extends BodyWithErrors, Valid
           return respond(503, request.body.value);
         } else if (request.body.value.notFound) {
           return respond(404, request.body.value);
+        } else if (request.body.value.conflict) {
+          return respond(409, request.body.value);
         } else {
           return await responseValidation.invalid({
             ...request,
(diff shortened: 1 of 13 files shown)


src/back-end/lib/server/index.ts
```diff
@@ -423,6 +423,8 @@ export function wrapRespond<ValidReqB, InvalidReqB extends BodyWithErrors, Valid
           return respond(503, request.body.value);
         } else if (request.body.value.notFound) {
           return respond(404, request.body.value);
+        } else if (request.body.value.conflict) {
+          return respond(409, request.body.value);
```
dhruvio: 👍

src/migrations/tasks/20200213210722_cwu_subscribers.ts
```diff
@@ -0,0 +1,20 @@
...
+import Knex from 'knex';
+
+const logger = makeDomainLogger(consoleAdapter, 'migrations');
+
+export async function up(connection: Knex): Promise<void> {
+  await connection.schema.createTable('cwuOpportunitySubscribers', table => {
+    table.uuid('opportunity').references('id').inTable('cwuOpportunities').notNullable();
+    table.uuid('user').references('id').inTable('users').notNullable();
+    table.timestamp('createdAt').notNullable();
+    table.primary(['opportunity', 'user']);
```
dhruvio: 👍

PR #50, opened 2020-02-20
DM-282 allow creation of proposals with draft/submitted status

sutherlanda:
This PR modifies the create endpoint to allow draft and submitted proposals to be created.

Also included is a fix for a bug that occurs when awarding a proposal and updating all other proposals to not awarded.

### DIFF
--- a/src/back-end/lib/db.ts
+++ b/src/back-end/lib/db.ts
@@ -887,6 +887,14 @@ export const readOneCWUOpportunity = tryDb<[Id, Session], CWUOpportunity | null>
       .select('id')).map(row => row.id);
   }
 
+  // Add on subscription flag, if authenticated user
+  if (result && session.user) {
+    const subscription = await connection<RawCWUOpportunitySubscriber>('cwuOpportunitySubscribers')
+      .where({ opportunity: result.id, user: session.user.id })
+      .first();
+    result.subscribed = !!subscription;
+  }
+
   // If admin/owner, add on list of status histories
   if (result && session.user && (session.user.type === UserType.Admin || result.createdBy === session.user.id)) {
     const rawStatusArray = await connection<RawCWUOpportunityStatusRecord>('cwuOpportunityStatuses')
--- a/src/back-end/lib/db.ts
+++ b/src/back-end/lib/db.ts
@@ -9,7 +9,7 @@ import { Affiliation, AffiliationSlim, MembershipStatus, MembershipType } from '
 import { FileBlob, FilePermissions, FileRecord } from 'shared/lib/resources/file';
 import { CreateCWUOpportunityStatus, CWUOpportunity, CWUOpportunitySlim, CWUOpportunityStatus, CWUOpportunityStatusRecord, privateOpportunitiesStatuses, publicOpportunityStatuses } from 'shared/lib/resources/opportunity/code-with-us';
 import { Organization, OrganizationSlim } from 'shared/lib/resources/organization';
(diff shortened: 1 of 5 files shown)


src/back-end/lib/resources/proposal/code-with-us.ts
```diff
@@ -149,10 +158,28 @@ const resource: Resource = {
...
+            session: request.session,
+            opportunity: validatedCWUOpportunity.value.id,
+            status: validatedStatus.value,
+            attachments: validatedAttachments.value
+          });
+        }
+
         const validatedProposalText = proposalValidation.validateProposalText(proposalText);
         const validatedAdditionalComments = proposalValidation.validateAdditionalComments(additionalComments);
         const validatedProponent = await validateProponent(connection, proponent);
```
dhruvio: I think proponents need to be validated somehow for drafts since they are a nested structure?

dhruvio: I think it's type-checking for drafts now because `get(body, 'proponent')` in `parseRequestBody` is returning `any`, which will likely lead to a bug.

src/shared/lib/resources/proposal/code-with-us.ts
```diff
@@ -61,12 +75,17 @@ export type CreateProponentRequestBody
```
dhruvio: I suggest adding a function to `parseCreateProponentRequestBody(raw: any): CreateProponentRequestBody | null`, and calling the function in the back-end CRUD resource when creating.

src/back-end/lib/resources/proposal/code-with-us.ts
```diff
@@ -44,7 +45,7 @@ type ValidatedDeleteRequestBody = Id;
 type Resource = crud.Resource<
   SupportedRequestBodies,
   SupportedResponseBodies,
-  CreateRequestBody,
+  Omit<CreateRequestBody, 'status'> & { status: string },
```
dhruvio: You'll need to update this to:

```ts
Omit<CreateRequestBody, 'status' | 'proponent'> & { status: string; proponent: CreateProponentRequestBody | null; }
```

src/back-end/lib/db.ts
```diff
@@ -1591,9 +1592,11 @@ export const awardCWUProposal = tryDb<[Id, string, AuthenticatedSession], CWUPro
...
     await connection('cwuProposalStatuses')
       .transacting(trx)
-      .whereIn('proposal', async () => {
-        await connection('cwuProposals')
-          .where({ opportunity: proposalRecord?.opportunity });
+      .whereIn('proposal', async function() {
+        this
+          .select('id')
+          .from('cwuProposals')
+          .where({ opportunity: proposalRecord.opportunity });
```
dhruvio: Out of curiosity, why was this change made?

sutherlanda: The original wasn't working.  On Telegram, Jesse pointed out it was throwing an error when there were other proposals to mark as not awarded.  I needed to use `this` instead of referring to the `connection` instance.

dhruvio: ah got it, thanks!

src/shared/lib/resources/proposal/code-with-us.ts
sutherlanda: I've implemented this in the back-end resource now (except using a blank individual proponent instead of null as discussed).  Let me know if this works, happy to adjust some more if necessary.

PR #53, opened 2020-02-22
DM-287 add additional file permission check

sutherlanda:
This PR adds an additional file permissions check (after fine-grained file permissions check) based on the access to an opportunity or proposal that the specified file may be an attachment for.

Also fix various attachment related bugs in database and permissions layer.


src/back-end/lib/db.ts
```diff
@@ -668,6 +669,33 @@ export async function hasFilePermission(connection: Connection, session: Session
...
+
+  // If file is an attachment on a proposal, and requesting user has access to the proposal, allow
+  if (session.user) {
+    const proposals = await connection('cwuProposalAttachments as attachments')
+      .innerJoin('cwuProposals as proposals', 'proposals.id', '=', 'attachments.proposal')
+      .where({ 'attachments.file': id })
+      .select<RawCWUProposal[]>('proposals.*');
+
+    if (proposals.length > 0) {
+      return proposals.some(async proposal => await hasReadPermissionCWUProposal(connection, session, proposal.opportunity, proposal.id));
```
dhruvio: Does `.some` work with `async/await` ?

dhruvio: If not, you'll need to use a `for...of` loop

src/back-end/lib/db.ts
```diff
@@ -1303,9 +1331,9 @@ export const readOneCWUProposal = tryDb<[Id, Session], CWUProposal | null>(async
     .first();
 
   if (result) {
-    result.attachments = (await connection<{ id: Id }>('cwuProposalAttachments')
+    result.attachments = (await connection<{ file: Id }>('cwuProposalAttachments')
       .where({ proposal: result.id })
-      .select('file')).map(row => row.id);
+      .select('file')).map(row => row.file);
```
dhruvio: This is what I meant earlier about I/O in TypeScript not really being helpful. Because the `knex` functions return an `any`, our type-casting doesn't add any safety.

dhruvio: The type annotation is really just for documentation 🤷‍♂

src/back-end/lib/db.ts
dhruvio: Smart use of the `permissions` function btw

dhruvio:
Also, could you please merge `feature/dm-207-8/cwu-proposal-edit-view` into this branch and ensure there aren't any conflicts? Then, once merged, change this PR's base to `feature/dm-207-8/cwu-proposal-edit-view`. That branch has all the validation improvements we talked about over IM.

src/back-end/lib/db.ts
sutherlanda: Yeah, good catch, it's not clear that will always behave.  I've switched it to a for...of for clarity.

src/back-end/lib/db.ts
sutherlanda: Yeah it helps for inserts and updates, but doesn't do much for selects :/

sutherlanda:
> Also, could you please merge `feature/dm-207-8/cwu-proposal-edit-view` into this branch and ensure there aren't any conflicts? Then, once merged, change this PR's base to `feature/dm-207-8/cwu-proposal-edit-view`. That branch has all the validation improvements we talked about over IM.

Done! Thanks for the review.

src/back-end/lib/db.ts
dhruvio: Generally, built-in iterative functions with callbacks are synchronous and don't support async/await. If you need to iterate over a list and run async code, you should use either map & Promise.all, or a `for...of` loop.

PR #54, opened 2020-02-23
DM-295 add unique constraint on opportunity and proposal status tables

sutherlanda:


dhruvio:
Merged this into the branch for #53 since it included that PR's changes.

PR #56, opened 2020-02-24
DM-108 email notification module

sutherlanda:


sutherlanda:
@dhruvio this is ready for review.  I've put in stub content for the email notifications, but the hooks are in place.

PR #59, opened 2020-02-26
DM-342 add rank to return proposals

sutherlanda:
Adds a ranking to each proposal response (only for admins/opportunity authors).

Also simplifies calculation of most recent status on returned proposal.


src/back-end/lib/db.ts
```diff
@@ -1270,14 +1270,6 @@ async function rawCWUProposalSlimToCWUProposalSlim(connection: Connection, raw:
 export const readManyCWUProposals = tryDb<[Id], CWUProposalSlim[]>(async (connection, id) => {
   const results = await connection<RawCWUProposalSlim>('cwuProposals as prop')
     .where({ opportunity: id })
-    .join<RawCWUProposalSlim>('cwuProposalStatuses as stat', function() {
```
sutherlanda: @dhruvio I've moved this logic out of the join and into a separate query below.  I noticed that since I added the `.whereNot` clause for the event changes in a previous PR, that this join was not returning all results.

src/back-end/lib/permissions.ts
```diff
@@ -165,6 +165,10 @@ export async function readOneCWUProposal(connection: Connection, session: Sessio
         (session.user && await isCWUProposalAuthor(connection, session.user, proposalId)) || false;
 }
 
+export async function canSeeCWUProposalScore(connection: Connection, session: Session, opportunityId: string): Promise<boolean> {
+  return isAdmin(session) || (session.user && await isCWUOpportunityAuthor(connection, session.user, opportunityId) || false);
+}
+
```
dhruvio: This function's name and implementation are misleading, as a vendor can see their own proposals' scores and ranks once they have been awarded/notAwarded. Can you please update accordingly?

src/back-end/lib/db.ts
```diff
@@ -1384,6 +1406,16 @@ export const readOneCWUProposal = tryDb<[Id, Session], CWUProposal | null>(async
       .orderBy('createdAt', 'desc');
 
     result.history = await Promise.all(rawProposalStasuses.map(async raw => await rawCWUProposalHistoryRecordToCWUProposalHistoryRecord(connection, session, raw)));
+
+    if (isRankableCWUProposalStatus(result.status) && await canSeeCWUProposalScore(connection, session, result.opportunity)) {
```
dhruvio: See comment in permissions.ts below.

src/back-end/lib/db.ts
```diff
@@ -1286,14 +1278,40 @@ export const readManyCWUProposals = tryDb<[Id], CWUProposalSlim[]>(async (connec
...
+    proposal.status = statusResult.status;
+  }
+
+  // Read ranks for rankable proposals and apply to existing result set
+  const rankableProposals = results.filter(result => isRankableCWUProposalStatus(result.status));
+  const ranks = await connection
+    .from('cwuProposals')
+    .whereIn('id', rankableProposals.map(r => r.id))
+    .andWhere({ opportunity: id })
+    .select(connection.raw('id, RANK () OVER (ORDER BY score DESC) rank'));
```
dhruvio: No score/permissions check because the permissions check is done earlier, I presume?

dhruvio:
Changed the base to `development` as I merged the front-end branch.

src/back-end/lib/db.ts
sutherlanda: Yeah that's right, the permissions check is done at the readMany resource level

src/back-end/lib/permissions.ts
sutherlanda: I wasn't aware vendors would be able to see their scores at any point (this is different than the original devex).  I will update the function.

sutherlanda: Okay I've updated the function: `readCWUProposalScore` and it allows vendors to view their score and rank if the proposal status is in Awarded/NotAwarded (or if the user is Admin/Opp Author).

I also realized we weren't restricting score based on user type/proposal status, so I've added a check there as well (I think you added another ticket for this).

PR #60, opened 2020-02-28
DM-343 add mailer config validation

sutherlanda:


sutherlanda:
> Missing dev and prod validation for the mailer config (see concierge repo's config file)

Added

PR #61, opened 2020-02-28
DM-353 switch update to insert for not awarded status

sutherlanda:


src/back-end/lib/db.ts
```diff
@@ -1768,19 +1768,24 @@ export const awardCWUProposal = tryDb<[Id, string, AuthenticatedSession], CWUPro
...
+        .transacting(trx)
+        .insert({
+          id: generateUuid(),
+          proposal: id,
+          createdAt: now,
+          createdBy: session.user.id,
+          status: CWUProposalStatus.NotAwarded,
+          note: ''
+        });
+    }
```
dhruvio: We only want to add the NotAwarded status to other proposals that are either Evaluated or Awarded.

sutherlanda: Updated

PR #63, opened 2020-02-28
DM-350 all vendors to readMany their own proposals

sutherlanda:
Allow vendors to readMany their own proposals on an opportunity.

Also limits scores to only be included if the requesting user is an admin/gov

### DIFF
--- a/src/back-end/lib/db.ts
+++ b/src/back-end/lib/db.ts
@@ -1267,8 +1267,8 @@ async function rawCWUProposalSlimToCWUProposalSlim(connection: Connection, raw: 
   };
 }
 
-export const readManyCWUProposals = tryDb<[Id], CWUProposalSlim[]>(async (connection, id) => {
-  const results = await connection<RawCWUProposalSlim>('cwuProposals as prop')
+export const readManyCWUProposals = tryDb<[Session, Id], CWUProposalSlim[]>(async (connection, session, id) => {
+  const query = connection<RawCWUProposalSlim>('cwuProposals as prop')
     .where({ opportunity: id })
     .select<RawCWUProposalSlim[]>(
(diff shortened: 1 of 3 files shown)


src/back-end/lib/db.ts
```diff
@@ -1267,8 +1267,8 @@ async function rawCWUProposalSlimToCWUProposalSlim(connection: Connection, raw:
   };
 }
 
-export const readManyCWUProposals = tryDb<[Id], CWUProposalSlim[]>(async (connection, id) => {
-  const results = await connection<RawCWUProposalSlim>('cwuProposals as prop')
+export const readManyCWUProposals = tryDb<[Session, Id], CWUProposalSlim[]>(async (connection, session, id) => {
```
dhruvio: Can you change `Session` to `AuthenticatedSession` since anon users won't be reading proposals?

src/back-end/lib/resources/proposal/code-with-us.ts
```diff
@@ -100,7 +100,7 @@ const resource: Resource = {
       if (!await permissions.readManyCWUProposals(connection, request.session, request.query.opportunity)) {
         return respond(401, [permissions.ERROR_MESSAGE]);
       }
-      const dbResult = await db.readManyCWUProposals(connection, request.query.opportunity);
+      const dbResult = await db.readManyCWUProposals(connection, request.session, request.query.opportunity);
```
dhruvio: You can use `permissions.isSignedIn(request.session)` for the type system to know whether the session is an `AuthenticatedSession` or not.

src/back-end/lib/db.ts
sutherlanda: updated

PR #64, opened 2020-02-28
DM-351 limit draft/submitted proposals for admin/gov

sutherlanda:
Added additional checks to permissions module.  In cases where it's not possible to entirely restrict in permissions.ts, additional checks were added at the db level.

### DIFF
--- a/src/back-end/lib/db.ts
+++ b/src/back-end/lib/db.ts
@@ -10,7 +10,7 @@ import { Affiliation, AffiliationSlim, MembershipStatus, MembershipType } from '
 import { FileBlob, FilePermissions, FileRecord } from 'shared/lib/resources/file';
 import { CreateCWUOpportunityStatus, CWUOpportunity, CWUOpportunityEvent, CWUOpportunityHistoryRecord, CWUOpportunitySlim, CWUOpportunityStatus, privateOpportunitiesStatuses, publicOpportunityStatuses } from 'shared/lib/resources/opportunity/code-with-us';
 import { Organization, OrganizationSlim } from 'shared/lib/resources/organization';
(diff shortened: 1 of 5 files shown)


src/back-end/lib/db.ts
```diff
@@ -691,14 +691,15 @@ export async function hasAttachmentPermission(connection: Connection, session: S
...
+    const rawProposals = await connection('cwuProposalAttachments as attachments')
       .innerJoin('cwuProposals as proposals', 'proposals.id', '=', 'attachments.proposal')
       .where({ 'attachments.file': id })
       .select<RawCWUProposal[]>('proposals.*');
 
-    if (proposals.length > 0) {
-      for (const proposal of proposals) {
-        if (await hasReadPermissionCWUProposal(connection, session, proposal.opportunity, proposal.id)) {
+    if (rawProposals.length > 0) {
+      for (const rawProposal of rawProposals) {
```
dhruvio: You don't need the `if` condition here, if there aren't any rawProposals, the loop won't run.

src/back-end/lib/db.ts
```diff
@@ -1299,6 +1309,11 @@ export const readManyCWUProposals = tryDb<[Id], CWUProposalSlim[]>(async (connec
     proposal.status = statusResult.status;
   }
 
+  // Filter out any proposals not in UNDER_REVIEW or later status if admin/gov owner
+  if (session.user && session.user.type !== UserType.Vendor) {
+    results = results.filter(result => ![CWUProposalStatus.Draft, CWUProposalStatus.Submitted].includes(result.status));
```
dhruvio: This proposal status check should be abstracted to a simple helper function in the shared proposal resource since it's being used multiple times.

dhruvio: e.g. `isProposalVisibleToGovernment(proposal: CWUProposal): boolean)`

PR #66, opened 2020-03-02
DM-231 add automated CWU update hook

sutherlanda:
Added a before hook on all CRUD resource endpoints in the back-end that updates published CWU opportunities with lapsed proposal deadlines, and also updates the submitted proposals attached to those opportunities.

The hook update function is throttled to fire no more than once per minute (leading and trailing edges).  The throttling interval can be changed via configuration.

Also added option for enabling Knex debug mode, which can help with debugging some of the more complex queries during development.

** This PR has a db migration that will need to be run. **


src/back-end/lib/db.ts
```diff
@@ -1921,3 +1921,74 @@ export const readManyCWUSubscribedUsers = tryDb<[Id], User[]>(async (connection,
...
+          .on('opportunities.id', '=', 'versions.opportunity')
+          .andOn('versions.createdAt', '=',
+            connection.raw('(select max("createdAt") from "cwuOpportunityVersions" as versions2 where \
+              versions2.opportunity = opportunities.id)'));
+      })
+      .where({
+        'statuses.status': CWUOpportunityStatus.Published
+      })
+      .andWhere('versions.proposalDeadline', '<=', now)
+      .select<Array<{ id: Id }>>('opportunities.id'))?.map(result => result.id) || [];
```
dhruvio: What an adrenaline rush! 😅

sutherlanda: :) I might try to abstract some of those joins into functions when I do the refactoring/splitting of `db.ts`.

PR #69, opened 2020-03-04
DM-228 reporting metrics data model and resource

sutherlanda:
Implements a counter that can be used to track views on opportunities, etc.

### DIFF
--- a/src/back-end/lib/db/counter.ts
+++ b/src/back-end/lib/db/counter.ts
@@ -1,0 +1,66 @@
+import { tryDb } from 'back-end/lib/db';
+import { valid } from 'shared/lib/http';
+
+interface ViewCounter {
+  name: string;
+  count: number;
+}
+
+export const incrementCounters = tryDb<[string[]], Record<string, number>>(async (connection, names) => {
+  // Update existing counters
+  const existingCounters = await connection<ViewCounter>('viewCounters')
+    .whereIn('name', names)
+    .increment('count')
+    .update({}, 'name');
+
+  // Create new counters where applicable
+  for (const name of names) {
+    if (!existingCounters.includes(name)) {
+      await connection<ViewCounter>('viewCounters')
+        .insert({
+          name,
+          count: 1
+        });
+    }
+  }
+
+  return await getCounters(connection, names);
+});
+
+export const decrementCounters = tryDb<[string[]], Record<string, number>>(async (connection, names) => {
+  // Update existing counters
+  const existingCounters = await connection<ViewCounter>('viewCounters')
+    .whereIn('name', names)
+    .decrement('count')
+    .update({}, 'name');
+
+  // Create new counters where applicable
+  for (const name of names) {
+    if (!existingCounters.includes(name)) {
+      await connection<ViewCounter>('viewCounters')
+        .insert({
+          name,
+          count: 0
+        });
+    }
+  }
+
+  return await getCounters(connection, names);
+});
+
+export const getCounters = tryDb<[string[]], Record<string, number>>(async (connection, names) => {
+  let query = connection<ViewCounter>('viewCounters')
+    .select('*');
+
+  if (names.length > 0) {
+    query = query.whereIn('name', names);
+  }
+
+  const counters = await query;
+
+  const record: Record<string, number> = {};
+  for (const counter of counters) {
+    record[counter.name] = counter.count;
+  }
+  return valid(record);
+});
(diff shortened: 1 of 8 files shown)


src/back-end/lib/db/counter.ts
```diff
@@ -0,0 +1,66 @@
+import { tryDb } from 'back-end/lib/db';
+import { valid } from 'shared/lib/http';
+
+interface ViewCounter {
+  name: string;
+  count: number;
+}
```
dhruvio: Rename to `Counter` and move to shared counter resource

src/back-end/lib/permissions.ts
```diff
@@ -204,3 +204,9 @@ export async function editCWUProposal(connection: Connection, session: Session,
 export async function deleteCWUProposal(connection: Connection, session: Session, proposalId: string): Promise<boolean> {
   return session.user && await isCWUProposalAuthor(connection, session.user, proposalId) || false;
 }
+
+// Metrics.
+
+export function readAllCounters(session: Session): boolean {
+  return isAdmin(session);
+}
```
dhruvio: I may not have explained this in the ticket, so my apologies if that's the case. Can we add a separate function called `readManyCounters`, and allow all admins & gov users to do so? This permissions check should happen at the beginning of the `readMany` CRUD actions. Thinking is that non-admin CWU gov opp owners will need to see an opportunity's total views.

src/back-end/lib/db/counter.ts
```diff
@@ -0,0 +1,66 @@
...
+  // Create new counters where applicable
+  for (const name of names) {
+    if (!existingCounters.includes(name)) {
+      await connection<ViewCounter>('viewCounters')
+        .insert({
+          name,
+          count: 0
+        });
+    }
+  }
```
dhruvio: Doesn't look like this is used?

src/back-end/lib/db/counter.ts
```diff
@@ -0,0 +1,66 @@
...
+  name: string;
+  count: number;
+}
+
+export const incrementCounters = tryDb<[string[]], Record<string, number>>(async (connection, names) => {
+  // Update existing counters
+  const existingCounters = await connection<ViewCounter>('viewCounters')
+    .whereIn('name', names)
+    .increment('count')
+    .update({}, 'name');
```
dhruvio: Strange that knex requires you to call increment then update.

src/shared/lib/resources/counter.ts
```diff
@@ -0,0 +1,5 @@
+import { BodyWithErrors } from 'shared/lib/types';
+
+export interface UpdateValidationErrors extends BodyWithErrors {
+  name?: string[];
+}
```
dhruvio: Can you add `export type UpdateRequestBody = null;` for completeness? Front-end will use it.

src/back-end/lib/db/counter.ts
sutherlanda: No, but the ticket specified it, so I assumed it was for future use.  Should I remove?

src/back-end/lib/db/counter.ts
sutherlanda: It doesn't require it, but increment only returns # of incremented rows.  I use update here to get the names of the updated rows so I know which counters needed to be inserted in the next step.

src/back-end/lib/permissions.ts
sutherlanda: No problem, I will add this.  I'll take opportunityId from the name of the counter provided to determine whether a gov user is an owner of the opportunity corresponding to the counter.

sutherlanda: Actually, I realized that you probably meant all gov owners should be able to see all counters, not just the ones they own?

src/back-end/lib/resources/counter.ts
```diff
@@ -0,0 +1,90 @@
...
+
+type Resource = crud.Resource<
+  SupportedRequestBodies,
+  SupportedResponseBodies,
+  null,
+  null,
+  null,
+  null,
+  null,
+  null,
```
dhruvio: Need to use `UpdateRequestBody` here (I *think* it's this line)

src/back-end/lib/db/counter.ts
dhruvio: Yes please, the intention is just to increment counters

src/back-end/lib/permissions.ts
dhruvio: Yeah, no need to do extra opp-based validation. Any gov user can view any counter.

sutherlanda:
> One tiny change left, then we're good to merge.

Cool, updates done.

PR #70, opened 2020-03-04
DM-301 migrations for swu opportunities

sutherlanda:
WIP until data model has been reviewed


sutherlanda:
@dhruvio ready for review.  Thx

sutherlanda:
Realized there are a few items missing around phase validation here.  Adding WIP back for now.

src/shared/lib/validation/opportunity/sprint-with-us.ts
```diff
@@ -0,0 +1,274 @@
+import { uniq } from 'lodash';
+import { dateToMidnight, getNumber, getString, getStringArray, isDateInThePast } from 'shared/lib';
+import SKILLS from 'shared/lib/data/skills';
```
dhruvio: This module was originally intended for the CWU skills dropdown. Can you duplicate this module as `capabilities` and import that instead?

src/shared/lib/validation/opportunity/sprint-with-us.ts
```diff
@@ -0,0 +1,274 @@
...
+export function validateSWUOpportunityPhaseMaxBudget(raw: number, maxTotalBudget?: number): Validation<number> {
+  return validateNumber(raw, 0, maxTotalBudget);
+}
+
+export function validateCapability(raw: string): Validation<string> {
+  return SKILLS.includes(raw) ? valid(raw) : invalid(['Please select a skill from the list.']);
+}
+
+export function validateSWUOpportunityPhaseRequiredCapabilities(raw: string[]): ArrayValidation<string> {
+  if (!raw.length) { return invalid([['Please select at least one skill.']]); }
```
dhruvio: Change `skill` to `capability`

src/shared/lib/resources/opportunity/sprint-with-us.ts
```diff
@@ -0,0 +1,217 @@
...
+  teamQuestions?: CreateSWUTeamQuestionValidationErrors[];
+  attachments?: string[][];
+}
+
+// Update.
+
+export type UpdateRequestBody
+  = ADT<'edit', UpdateEditRequestBody>
+  | ADT<'publish', string>
+  | ADT<'evaluateQuestions', string>
```
dhruvio: `evaluateQuestions` would only be called automatically by a hook when a proposal deadline lapses — no need to have it as part of the update request body.

src/shared/lib/resources/opportunity/sprint-with-us.ts
```diff
@@ -0,0 +1,217 @@
...
+  | ADT<'suspend', string>
+  | ADT<'cancel', string>
+  | ADT<'addAddendum', string>;
+
+export type UpdateEditRequestBody = Omit<CreateRequestBody, 'status'>;
+
+type UpdateADTErrors
+  = ADT<'edit', UpdateEditValidationErrors>
+  | ADT<'publish', string[]>
+  | ADT<'evaluateQuestions', string[]>
```
dhruvio: Remove

src/shared/lib/resources/opportunity/sprint-with-us.ts
```diff
@@ -0,0 +1,217 @@
...
+  updatedBy: UserSlim;
+}
+
+export type SWUOpportunitySlim = Pick<SWUOpportunity, 'id' | 'title' | 'createdAt' | 'createdBy' | 'updatedAt' | 'updatedBy' | 'status' | 'proposalDeadline'>;
+
+// Create.
+
+export type CreateSWUOpportunityStatus
+  = SWUOpportunityStatus.Published
+  | SWUOpportunityStatus.Draft;
```
dhruvio: What about submitting a SWU opp for review?

src/shared/lib/validation/opportunity/sprint-with-us.ts
```diff
@@ -0,0 +1,274 @@
...
+export function validateSWUOpportunityPhaseCompletionDate(raw: string, startDate: Date): Validation<Date> {
+  return validateDate(raw, startDate);
+}
+
+export function validateSWUOpportunityPhaseMaxBudget(raw: number, maxTotalBudget?: number): Validation<number> {
+  return validateNumber(raw, 0, maxTotalBudget);
+}
+
+export function validateCapability(raw: string): Validation<string> {
+  return SKILLS.includes(raw) ? valid(raw) : invalid(['Please select a skill from the list.']);
```
dhruvio: change `skill` to `capability`

src/migrations/tasks/20200305134738_swu_opportunities.ts
```diff
@@ -0,0 +1,170 @@
...
+    table.timestamp('updatedAt').notNullable();
+    table.uuid('updatedBy').references('id').inTable('users').notNullable();
+    table.primary(['phase', 'capability']);
+  });
+  logger.info('Created swuPhaseCapabilities table.');
+
+  await connection.schema.createTable('swuTeamQuestions', table => {
+    table.uuid('opportunityVersion').references('id').inTable('swuOpportunityVersions').notNullable().onDelete('CASCADE');
+    table.string('question', 1000).notNullable();
+    table.string('guideline', 1000).notNullable();
```
dhruvio: Are the `1000`s char limits? Don't add these in the DM, this should be validated using validation functions, easier to change as needed in the future.

sutherlanda: They are char limits, but knex defaults to 255 if you don't specify.  We may need more than that for those items.

src/migrations/tasks/20200305134738_swu_opportunities.ts
```diff
@@ -0,0 +1,170 @@
...
+    table.timestamp('updatedAt').notNullable();
+    table.uuid('updatedBy').references('id').inTable('users').notNullable();
+
+    table.unique(['opportunityVersion', 'phase']);
+  });
+  logger.info('Created swuOpportunityPhases table.');
+
+  await connection.schema.createTable('swuPhaseCapabilities', table => {
+    table.uuid('phase').references('id').inTable('swuOpportunityPhases').notNullable().onDelete('CASCADE');
+    table.enum('capability', SKILLS).notNullable();
```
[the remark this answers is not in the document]
sutherlanda: Sure

src/shared/lib/validation/opportunity/sprint-with-us.ts
sutherlanda: Sure.  I've also update the list to be consistent with the wireframes/designs.

src/shared/lib/resources/opportunity/sprint-with-us.ts
sutherlanda: Right, added.  Also updated the migration to allow this enum value.

src/shared/lib/resources/opportunity/sprint-with-us.ts
sutherlanda: Removed, and added 'submitForReview' update request body.

sutherlanda:
Thanks @dhruvio. Changes completed and is ready for re-review.

src/migrations/tasks/20200305134738_swu_opportunities.ts
dhruvio: In past migrations, we create quite a few string columns without specifying the char limit — these should probably be changed too? Also, does postgres support a "text" column type without the char limit?

dhruvio: Don't make the change yet — let's figure out a solution first

sutherlanda: We should be using text.  From what I'm seeing in the database, Knex with Postgres will still only allow 255 max despite specifying the length because it uses varchar, which has a hard limit of 255 characters.  Text does not have a limit.

I can do a migration to update our larger string fields to use text instead.  Thanks for catching this.

sutherlanda: Will issue new migration in a new PR as discussed.

PR #72, opened 2020-03-06
DM-305 data model for SWU proposals

sutherlanda:
WIP until data model/ER diagram for SWU proposals has been reviewed.

### DIFF
--- a/src/migrations/tasks/20200305215313_swu_proposals.ts
+++ b/src/migrations/tasks/20200305215313_swu_proposals.ts
@@ -76,6 +76,7 @@ export async function up(connection: Knex): Promise<void> {
     table.uuid('proposal').references('id').inTable('swuProposals').notNullable().onDelete('CASCADE');
     table.enu('phase', Object.values(SWUProposalPhaseType)).notNullable();
     table.integer('proposedCost').notNullable();
+    table.unique(['proposal', 'phase']);
   });
   logger.info('Created swuProposalPhases table.');
 
@@ -107,6 +108,12 @@ export async function up(connection: Knex): Promise<void> {
   logger.info('Created swuTeamQuestionResponses table.');
 }
 
-// tslint:disable-next-line: no-empty
 export async function down(connection: Knex): Promise<void> {
+  await connection.schema.dropTable('swuTeamQuestionResponses');
+  await connection.schema.dropTable('swuProposalReferences');
+  await connection.schema.dropTable('swuProposalTeamMembers');
+  await connection.schema.dropTable('swuProposalPhases');
+  await connection.schema.dropTable('swuProposalAttachments');
+  await connection.schema.dropTable('swuProposalStatuses');
+  await connection.schema.dropTable('swuProposals');
 }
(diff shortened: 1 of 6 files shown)


src/migrations/tasks/20200305215313_swu_proposals.ts
```diff
@@ -0,0 +1,112 @@
...
+    table.uuid('file').references('id').inTable('files').notNullable();
+    table.primary(['proposal', 'file']);
+  });
+
+  logger.info('Created swuProposalAttachments table.');
+
+  await connection.schema.createTable('swuProposalPhases', table => {
+    table.uuid('id').primary().unique().notNullable();
+    table.uuid('proposal').references('id').inTable('swuProposals').notNullable().onDelete('CASCADE');
+    table.enu('phase', Object.values(SWUProposalPhaseType)).notNullable();
```
dhruvio: Add unique constraint on proposal and phase columns

src/shared/lib/resources/proposal/sprint-with-us.ts
```diff
@@ -0,0 +1,194 @@
...
+  opportunity?: string[];
+  status?: string[];
+}
+
+// Update.
+
+export type UpdateRequestBody
+  = ADT<'edit', UpdateEditRequestBody>
+  | ADT<'submit', string>
+  | ADT<'score', string>
```
dhruvio: Need to support scoring questions, code challenge, and team scenario

src/shared/lib/resources/proposal/sprint-with-us.ts
```diff
@@ -0,0 +1,194 @@
...
+  | ADT<'award', string>
+  | ADT<'disqualify', string>
+  | ADT<'withraw', string>;
+
+export type UpdateEditRequestBody = Omit<CreateRequestBody, 'opportunity' | 'status'>;
+
+type UpdateADTErrors
+  = ADT<'edit', UpdateEditValidationErrors>
+  | ADT<'submit', string[]>
+  | ADT<'score', string[]>
```
dhruvio: See previous comment.

src/shared/lib/resources/proposal/sprint-with-us.ts
dhruvio: Not sure when pricing should be scored by the system? Once the proposal deadline lapses, after questions are scored, or after the scenario is scored? Are disqualified/screened out proposals included in price scoring?

sutherlanda: I think it will have to be after the code challenge/scenario is scored.  Disqualified/screened out proposals are not included in the price scoring (at least in the current devex).

I'll add ADTs for scoring questions, code challenge, and scenario, but let me know if you think we should modify this further.

sutherlanda:
Thanks @dhruvio , changes have been completed.

src/shared/lib/resources/proposal/sprint-with-us.ts
dhruvio: That works for me. Sounds good re: automatically scoring price once all qualified team scenarios have been scored.

PR #74, opened 2020-03-10
DM-310 add user capabilities data model/resource/validation

sutherlanda:


src/migrations/tasks/20200309204727_user_capabilities.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export async function up(connection: Knex): Promise<void> {
+  await connection.schema.alterTable('users', table => {
+    table.specificType('capabilities', 'TEXT[]').defaultTo('{}').notNullable();
+  });
+
+  await connection.schema.raw(`
+  ALTER TABLE "users" \
+  ADD CONSTRAINT "users_capabilities_check" \
+  CHECK (capabilities <@ ARRAY['${Object.values(SKILLS).join('\',\'')}']) \
```
dhruvio: IMO don't need to check the capability is valid in the DB, just use validation functions to do the check

src/shared/lib/resources/user.ts
```diff
@@ -69,17 +70,27 @@ export interface UpdateProfileRequestBody {
   avatarImageFile?: Id;
 }
 
+export interface UpdateCapabilitiesRequestBody {
+  capabilities: string[];
+}
```
dhruvio: Don't need this, see below

src/shared/lib/resources/user.ts
```diff
@@ -69,17 +70,27 @@ export interface UpdateProfileRequestBody {
...
+
 export type UpdateRequestBody
   = ADT<'updateProfile', UpdateProfileRequestBody>
+  | ADT<'updateCapabilities', UpdateCapabilitiesRequestBody>
```
dhruvio: The value can simply be an array of strings

src/shared/lib/resources/user.ts
```diff
@@ -69,17 +70,27 @@ export interface UpdateProfileRequestBody {
...
 
 export type UpdateProfileValidationErrors = ErrorTypeFrom<UpdateProfileRequestBody>;
 
+export interface UpdateCapabilitiesValidationErrors {
+  capabilities?: string[][];
+}
+
 type UpdateADTErrors
   = ADT<'updateProfile', UpdateProfileValidationErrors>
+  | ADT<'updateCapabilities', UpdateCapabilitiesValidationErrors>
```
dhruvio: The capability errors can simply be `string[][]`

src/shared/lib/validation/user.ts
```diff
@@ -1,5 +1,7 @@
+import { uniq } from 'lodash';
+import SKILLS from 'shared/lib/data/skills';
```
dhruvio: See comment from SWU opp PR — this module should be duplicated and named as `capabilities`

src/shared/lib/validation/user.ts
```diff
@@ -20,3 +22,12 @@ export function validateUserStatus(status: string): Validation<UserStatus> {
...
 }
+
+export function validateCapability(raw: string): Validation<string> {
+  return SKILLS.includes(raw) ? valid(raw) : invalid(['Invalid capability specified.']);
+}
+
+export function validateCapabilities(raw: string[]): ArrayValidation<string> {
+  const validatedArray = validateArray(raw, validateCapability);
+  return mapValid(validatedArray, v => uniq(v));
+}
```
dhruvio: This can be moved to `shared/lib/validation/index.ts` and re-exported from this module and the SWU opp validation module.

src/migrations/tasks/20200309204727_user_capabilities.ts
sutherlanda: Agreed, removed.

sutherlanda:
Thanks, ready for re-review @dhruvio

src/back-end/lib/resources/user.ts
```diff
@@ -158,17 +160,12 @@ const resource: Resource = {
...
-              return valid(adt('updateCapabilities' as const, {
-                capabilities: validatedCapabilities.value
-              }));
+              return valid(adt('updateCapabilities' as const, validatedCapabilities.value));
             } else {
               return invalid({
-                user: adt('updateCapabilities' as const, {
-                  capabilities: getInvalidValue(validatedCapabilities, undefined)
-                })
+                user: adt('updateCapabilities' as const, getInvalidValue(validatedCapabilities, undefined) as string[][])
```
dhruvio: 2nd arg to `getInvalidValue` should be an empty array, not undefined per the shared type.

sutherlanda: Thanks, corrected.

PR #75, opened 2020-03-11
DM-311 rfq swu qualified status / member lists

sutherlanda:
This PR includes the following:

1. Modified readMany affiliations to take an optional `organization` query param.  If included, this endpoint will return the team members for that org id (permission for admin/org owner only)
2. Added an `acceptedSWUTerms` field to the Organization data model and type.  It's a nullable date field.
3. Added a `swuQualified` boolean flag to the Organization type.  This is calculated and included on readOne and readMany responses for admins/org owners only (similar to owner information).  It is set  true if there are at least 2 active members, all capabilities across the team are met, and the terms have been accepted.

The one piece left to do is how to accept terms.  I can do this as an update on the organization with an ADT to separate regular org profile updates from term acceptance, or just include it as part of the existing org update.  @dhruvio let me know what's preferred, as the first option would require work on the frontend.

### DIFF
--- a/src/back-end/lib/db/organization.ts
+++ b/src/back-end/lib/db/organization.ts
@@ -1,7 +1,9 @@
+import CAPABILITIES from 'back-end/../shared/lib/data/capabilities';
 import { generateUuid } from 'back-end/lib';
 import { Connection, tryDb } from 'back-end/lib/db';
-import { createAffiliation } from 'back-end/lib/db/affiliation';
+import { createAffiliation, readManyAffiliationsForOrganization } from 'back-end/lib/db/affiliation';
 import { readOneFileById } from 'back-end/lib/db/file';
+import { spread, union } from 'lodash';
 import { valid } from 'shared/lib/http';
 import { MembershipStatus, MembershipType } from 'shared/lib/resources/affiliation';
 import { FileRecord } from 'shared/lib/resources/file';
@@ -25,6 +27,7 @@ interface RawOrganizationSlim extends Omit<OrganizationSlim, 'logoImageFile' | '
   logoImageFile?: Id;
   ownerId?: Id;
   ownerName?: string;
+  acceptedSWUTerms?: boolean;
 }
 
 async function rawOrganizationToOrganization(connection: Connection, raw: RawOrganization): Promise<Organization> {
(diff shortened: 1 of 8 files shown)


dhruvio:
Thanks for the explanation. Let's go with ADTs; cleaner. Thanks!

sutherlanda:
ADT for organization updates added.  Ready for review @dhruvio .  Thanks.

src/back-end/lib/db/organization.ts
```diff
@@ -66,6 +69,21 @@ async function rawOrganizationSlimToOrganizationSlim(connection: Connection, raw
...
 
+async function calculateSWUQualifiedStatus(connection: Connection, organization: RawOrganization | RawOrganizationSlim): Promise<boolean> {
+  const dbResult = await readManyAffiliationsForOrganization(connection, organization.id);
+  if (isInvalid(dbResult) || !dbResult.value) {
+    return false;
+  }
+
+  // Need at least two ACTIVE members, all capabilities between members, and accepted terms
+  const activeMembers = dbResult.value.filter(a => a.membershipStatus === MembershipStatus.Active);
+  const unionedCapabilities = spread<string[]>(union)(activeMembers.map(m => m.user.capabilities));
```
dhruvio: 👍

src/shared/lib/resources/affiliation.ts
```diff
@@ -26,8 +26,15 @@ export interface Affiliation {
...
+  membershipType: MembershipType;
   organization: Pick<Organization, 'id' | 'legalName'>;
+}
+
+export interface AffiliationMember {
+  id: Id;
   membershipType: MembershipType;
+  membershipStatus: MembershipStatus;
+  user: Pick<User, 'id' | 'name' | 'avatarImageFile' | 'capabilities'>;
 }
```
dhruvio: I think you can use `AffiliationMember` as the new `AffiliationSlim` and avoid needing two separate types for the read many action in the affiliation resource. Unless I'm missing something?

sutherlanda: I did it this way because `AffiliationSlim` includes the org, so `rawAffiliationToAffiliationSlim` uses `readOneOrganization`, which in turn invokes `calculateSWUQualifiedStatus` causing a loop.

`AffiliationSlim` is used for getting a list of orgs a user belongs to so needs to include org.
`AffiliationMember` is used for getting a list of users belonging to an org so does not need to include org.

dhruvio: Got it — thanks for the explanation

PR #77, opened 2020-03-12
DM-368 Update data fields using string to text where required

sutherlanda:


dhruvio:
IMO, best to use text wherever possible and avoid using the string field.

sutherlanda:
OK

sutherlanda:
Ready for re-review @dhruvio

PR #81, opened 2020-03-18
DM-314 resource for SWU opportunities

sutherlanda:


src/shared/lib/validation/opportunity/sprint-with-us.ts
```diff
@@ -76,8 +77,8 @@ export function validatePhaseRequiredCapability(raw: any): Validation<CreateSWUO
 }
 
 export function validateSWUOpportunityInceptionPhase(raw: any, opportunityAssignmentDate: Date): Validation<ValidatedCreateSWUOpportunityPhaseBody, CreateSWUOpportunityPhaseValidationErrors> {
-  const validatedStartDate = validateSWUOpportunityInceptionPhaseStartDate(getString(raw, 'startDate'), opportunityAssignmentDate);
-  const validatedCompletionDate = validateSWUOpportunityPhaseCompletionDate(getString(raw, 'completionDate'), getValidValue(validatedStartDate, new Date()));
+  const validatedStartDate = validateSWUOpportunityInceptionPhaseStartDate(getISODateString(raw, 'startDate'), opportunityAssignmentDate);
+  const validatedCompletionDate = optional(getISODateString(raw, 'completionDate'), v => validateSWUOpportunityPhaseCompletionDate(getString(v, 'completionDate'), getValidValue(validatedStartDate, new Date())));
```
dhruvio, line 81: I don't think you're meant to use `getString` here

src/shared/lib/validation/opportunity/sprint-with-us.ts
```diff
@@ -76,8 +77,8 @@ export function validatePhaseRequiredCapability(raw: any): Validation<CreateSWUO
 }
 
 export function validateSWUOpportunityInceptionPhase(raw: any, opportunityAssignmentDate: Date): Validation<ValidatedCreateSWUOpportunityPhaseBody, CreateSWUOpportunityPhaseValidationErrors> {
```
dhruvio, line 79: Are the arguments to these validation functions meant to be the entire SWU opp object? Can't use them in the front-end if that's the case -- much easier if the `raw` value is the thing to be validated. In this case, that would be the `completionDate`. Could you review the other validation functions if that's the case? Let me know if you want to discuss.

sutherlanda: So that validation function was meant for validating a phase body on the SWU opp create body.  That's the raw in this case.  I pass in the opportunityAssignmentDate because it's needed to validate the start date for the phase.  I might be misunderstanding.

sutherlanda:
@dhruvio, this branch should be good to merge now if you're happy with it.  I'll do the rest of the changes we discussed today in the swu-proposals PR.

PR #83, opened 2020-03-22
DM-313 swu proposal resource

sutherlanda:
Back-end resource and db routines for SWU proposals


sutherlanda:
@dhruvio this is ready for another review

src/shared/lib/resources/proposal/sprint-with-us.ts
```diff
@@ -56,18 +64,19 @@ export interface SWUProposal {
...
   submittedAt?: Date;
   opportunity: SWUOpportunitySlim;
-  organization: OrganizationSlim;
+  organization?: OrganizationSlim;
   inceptionPhase?: SWUProposalPhase;
   prototypePhase?: SWUProposalPhase;
-  implementationPhase: SWUProposalPhase;
-  references: SWUProposalReference[];
-  attachments: FileRecord[];
+  implementationPhase?: SWUProposalPhase;
```
dhruvio, line 70: Could you remind me -- why did this become optional?

sutherlanda: For the anonymized proposals - we can't return any phase information as it would contain team members, which could be used to identify the proposal.

PR #84, opened 2020-03-30
DM-369 populate reporting metrics and success proponent flag on opportunities

sutherlanda:


src/shared/lib/resources/opportunity/code-with-us.ts
```diff
@@ -51,10 +51,7 @@ export interface CWUOpportunity {
   createdBy?: UserSlim;
   updatedBy?: UserSlim;
 
-  // TODO
-  // @successful-proponent
-  successfulProponent?: true;
-
+  successfulProponent?: boolean;
```
dhruvio, line 54: My apologies -- I think I used boolean as a stub value. This should be:

`successfulProposal?: CWUProposalSlim;`

Same for SWU below.

sutherlanda:
> Looks great. I made a mistake with the stub `successfulProponent` property. It should be `successfulProposal`. See my comment below. Sorry! 🇨🇦
> 
> EDIT: I missed something else. We also need the `successfulProponentName?: string` property for the opp view pages. `successfulProposal` should only be visible to admins/opp owners, and `successfulProponentName` can be visible to everyone.

No problem! I'll make those changes today.

sutherlanda:
@dhruvio requested changes are complete.

PR #85, opened 2020-04-06
Squashed changes

jjbandit:
Howdy!

The tabbing situation is not integrated in the way the rest of the tabs in the app are.  I decided to do something that would work for me first, then integrate it with the framework later.  This integration work proved slow and frustrating, so it has been left unfinished.  The groundwork has been laid out for it to be completed; the route is setup correctly, and the empty components are in place and compiling.  The parts of the page that depend on that being done are also incomplete, specifically the context menu in the header.

I also was unsure about how the backend serves the data for organization team members, so I used a static array of dummy data to do the testing.

The commit history is squashed because when I was merging the development branch in I got conflicts in the same file(s) several times, and decided it would be faster to squash the commits and deal with all conflicts at the same time than over and over while rebasing to preserve history.

Cheers,
Jesse

### DIFF
--- a/src/front-end/sass/index.scss
+++ b/src/front-end/sass/index.scss
@@ -387,6 +387,13 @@ table {
     @extend .bg-light;
   }
 
+  /* Note(Jesse) : In order to have the table cells not 'pop' when content
+   * that's hidden gets its display property toggled, this can be set to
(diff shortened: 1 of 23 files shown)


dhruvio:
Thanks, Jesse. Can you please create a PR with the full commit log for your work?

PR #89, opened 2020-04-08
DM-423 convert 'pending' to calculated field

sutherlanda:
I've converted 'pending' on SWUProposalTeamMember to be a readonly/calculated field.  It's no longer persisted in the database.

Also added validation to check for multiple scrum masters per phase has been added.

I also discovered a bug while testing this change where saved proposals were not getting all of their phase values (proposed cost/team members) checked when submitted.   The last commit addresses that.


src/back-end/lib/db/proposal/sprint-with-us.ts
```diff
@@ -158,6 +159,28 @@ async function rawSWUProposalToSWUProposal(connection: Connection, session: Sess
...
+      .select('membershipStatus')
+      .first())?.membershipStatus === MembershipStatus.Pending;
+  }
+
+  for (const member of implementationPhase.members) {
+    member.pending = (await connection('affiliations')
+      .where({ user: member.member.id, organization: organizationId })
+      .select('membershipStatus')
+      .first())?.membershipStatus === MembershipStatus.Pending;
+  }
```
dhruvio, line 182: Will the use of `first` here incorrectly return `INACTIVE` affiliations if there are any?

sutherlanda: Yes, nice catch.  Updated with a more restrictive where clause.

PR #92, opened 2020-04-09
DM-425 fix attachment delete bug for CWU proposals

sutherlanda:
Bug was caused by a logic error in creating/updating proposals.  The offending routine has been simplified and refactored (same approach used with SWU).


src/back-end/lib/db/proposal/code-with-us.ts
```diff
@@ -124,27 +124,19 @@ async function getCWUProposalSubmittedAt(connection: Connection, proposal: RawCW
...
-        .insert({
-          proposal: proposalId,
-          file: attachment.id
-        }, '*');
-      if (!attachmentResult) {
-        throw new Error('Unable to create proposal attachment');
-      }
+      }, '*');
+    if (!attachmentResult) {
+      throw new Error('Unable to create proposal attachment');
```
dhruvio, line 139: :+1:

PR #95, opened 2020-04-12
DM-400 notification reference page

sutherlanda:
Also refactored notifications for cwu/swu/users.


src/back-end/config.ts
```diff
@@ -34,6 +34,8 @@ export const BASIC_AUTH_PASSWORD_HASH = get('BASIC_AUTH_PASSWORD_HASH', '');
 
 export const ORIGIN = get('ORIGIN', 'http://digital-marketplace.bcgov.realfolk.io').replace(/\/*$/, '');
 
+export const CONTACT_EMAIL = process.env.CONTACT_EMAIL || 'TODO';
```
dhruvio, line 37: You should use the locally-defined `get` function here

sutherlanda: Ah, right.  Updated.

PR #97, opened 2020-04-13
DM-385 add word count validation to SWU proposal

sutherlanda:
Also fix issue with reading team questions.


src/shared/lib/validation/proposal/sprint-with-us.ts
```diff
@@ -83,34 +85,47 @@ export function validateSWUProposalReferences(raw: any): ArrayValidation<CreateS
...
+    return invalid({
+      order: getInvalidValue(validatedOrder, undefined)
+    });
+  }
+  const wordLimit = opportunityTeamQuestions.find(q => q.order === validatedOrder.value)?.wordLimit || 0;
+  if (!wordLimit) {
+    return invalid({
+      order: ['No matching opportunity question.']
+    });
+  }
```
dhruvio, line 112: This block will evaluate if the `wordLimit` is zero. Not sure if that is intentional?

src/shared/lib/validation/proposal/sprint-with-us.ts
```diff
@@ -83,34 +85,47 @@ export function validateSWUProposalReferences(raw: any): ArrayValidation<CreateS
   return validateArrayCustom(raw, validateSWUProposalReference, {});
 }
 
-// TODO validate word limit
 export function validateSWUProposalTeamQuestionResponseResponse(raw: string, wordLimit = MAX_TEAM_QUESTION_WORD_LIMIT): Validation<string> {
-  return validateGenericString(raw, 'Response', 1);
+  if (count(raw, 'words', {}) > wordLimit) {
```
dhruvio, line 89: Also, could you move this to a helper like `countWords` in `shared/lib/index.ts`? Also, we should have a generic validation function like `validateGenericStringWords` to validate by word count instead of char count, and use it here.

src/shared/lib/validation/proposal/sprint-with-us.ts
sutherlanda: Yeah, I default it to 0 in the line above if it can't match the order up with the opportunity questions.  To make it clearer though, I should default to null.  Not sure why I picked 0.

src/shared/lib/validation/proposal/sprint-with-us.ts
sutherlanda: Will do.

dhruvio: TY

sutherlanda:
@dhruvio, request changes are complete :)

src/shared/lib/validation/index.ts
```diff
@@ -149,6 +155,15 @@ export function validateGenericString(value: string, name: string, min = 1, max
   }
 }
 
+export function validateGenericStringWords(value: string, name: string, min = 1, max = 3000, words = 'words'): Validation<string> {
+  const count = countWords(value);
+  if (count < 1 || count > max) {
```
dhruvio, line 160: Should compare `count` to `min`

PR #99, opened 2020-04-14
DM-438 add migration to prevent deadlocks

sutherlanda:
Modified the session cleanup trigger to include an UPDATE lock that prevents any of the affected records from being updated by other operations during execution.

### DIFF
--- a/src/migrations/tasks/20200413164227_fix-session-trigger-lock.ts
+++ b/src/migrations/tasks/20200413164227_fix-session-trigger-lock.ts
@@ -1,0 +1,76 @@
+import { makeDomainLogger } from 'back-end/lib/logger';
+import { console as consoleAdapter } from 'back-end/lib/logger/adapters';
+import Knex from 'knex';
+
+const logger = makeDomainLogger(consoleAdapter, 'migrations');
+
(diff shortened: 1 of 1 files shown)


src/migrations/tasks/20200413164227_fix-session-trigger-lock.ts
```diff
@@ -0,0 +1,76 @@
...
+  // The update lock will prevent updates to any of the affected rows during execution of this function.
+  await connection.schema.raw(`
+    CREATE FUNCTION delete_old_sessions() RETURNS trigger
+      LANGUAGE plpgsql
+      AS $$
+    DECLARE
+      rows TEXT;
+    BEGIN
+      SELECT id FROM sessions WHERE "createdAt" < NOW() - INTERVAL '7 days' FOR UPDATE INTO rows;
+      DELETE FROM sessions WHERE "createdAt" < NOW() - INTERVAL '7 days';
```
dhruvio, line 29: Out of curiosity, does this cause the original issue? Is it possible to `DELETE rows;`?

dhruvio: Resolved, no need to change for now.

PR #104, opened 2020-04-16
DM-414 limit withdrawn proposal visibility

sutherlanda:
Also includes some minor refactoring of CWU/SWU proposal queries

### DIFF
--- a/src/shared/lib/resources/proposal/sprint-with-us.ts
+++ b/src/shared/lib/resources/proposal/sprint-with-us.ts
@@ -261,6 +261,7 @@ export function isSWUProposalStatusVisibleToGovernment(s: SWUProposalStatus): bo
   switch (s) {
     case SWUProposalStatus.Draft:
     case SWUProposalStatus.Submitted:
+    case SWUProposalStatus.Withdrawn:
       return false;
     default:
       return true;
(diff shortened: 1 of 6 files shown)


sutherlanda:
> I should have clarified in the ticket. The req is to hide withdrawn proposals from Government user types (not Admins). Looks like this hides withdrawn proposals from both Government and Admins.
> 
> It's confusing jargon, when I say gov users, I mean of user type "Government", when I say "public sector" users I mean either Government or Admin user types.

Okay, updated.

src/shared/lib/resources/proposal/sprint-with-us.ts
```diff
@@ -257,12 +257,13 @@ export interface DeleteValidationErrors extends BodyWithErrors {
   status?: string[];
 }
 
-export function isSWUProposalStatusVisibleToGovernment(s: SWUProposalStatus): boolean {
+export function isSWUProposalStatusVisibleToGovernment(s: SWUProposalStatus, role: UserType): boolean {
```
dhruvio, line 260: Last minor change, please update the `UserType` type signature to a union of `UserType.Government | UserType.Admin` to exclude vendors.

dhruvio: Same for CWU

PR #105, opened 2020-04-16
DM-483 permissions update to SWU opps

sutherlanda:


src/back-end/lib/permissions.ts
```diff
@@ -227,8 +227,21 @@ export function publishSWUOpportunity(session: Session): boolean {
   return isAdmin(session);
 }
 
-export async function deleteSWUOpportunity(connection: Connection, session: Session, opportunityId: string): Promise<boolean> {
-  return isAdmin(session) || (session && isGovernment(session) && await isSWUOpportunityAuthor(connection, session.user, opportunityId)) || false;
+export async function deleteSWUOpportunity(connection: Connection, session: Session, opportunityId: string, status: SWUOpportunityStatus): Promise<boolean> {
+  return isAdmin(session) && [SWUOpportunityStatus.Draft, SWUOpportunityStatus.UnderReview].includes(status)
```
dhruvio, line 231: Can you please add parentheses around the admin check before the `||` operator for legibility?

src/back-end/lib/permissions.ts
```diff
@@ -227,8 +227,21 @@ export function publishSWUOpportunity(session: Session): boolean {
...
+    || (session && isGovernment(session) && await isSWUOpportunityAuthor(connection, session.user, opportunityId) && status === SWUOpportunityStatus.Draft) || false;
```
dhruvio, line 232: You don't need the `session &&` bit

src/back-end/lib/permissions.ts
sutherlanda: Done

src/back-end/lib/permissions.ts
sutherlanda: It throws a type error on the following call if I don't due to TypeScript weirdness.  I've changed it to use isSignedIn() for consistency.

PR #122, opened 2020-07-28
Sprint With Us Word template

jw173:
This Word template can be used to help draft Sprint With Us opportunities


dhruvio:
Thanks, @jw173. Will review.

PR #127, opened 2020-09-02
DM-653 add permissions check for terms accepted

sutherlanda:
This PR adds a permissions check on certain actions by Vendor users to ensure they have accepted the terms and conditions prior to:

* Creating an organization
* Creating/submitting a CWU proposal
* Creating/submitting a SWU proposal

I didn't add the check to editing/deleting a proposal as we already check for authorship on the proposal, which would imply they had permission to create it.

### DIFF
--- a/src/back-end/lib/permissions.ts
+++ b/src/back-end/lib/permissions.ts
@@ -1,4 +1,4 @@
-import { Connection, hasCWUAttachmentPermission, hasFilePermission, isCWUOpportunityAuthor, isCWUProposalAuthor, isSWUOpportunityAuthor, isUserOwnerOfOrg } from 'back-end/lib/db';
+import { Connection, hasCWUAttachmentPermission, hasFilePermission, isCWUOpportunityAuthor, isCWUProposalAuthor, isSWUOpportunityAuthor, isUserOwnerOfOrg, userHasAcceptedTerms } from 'back-end/lib/db';
 import { hasSWUAttachmentPermission, isSWUProposalAuthor } from 'back-end/lib/db/proposal/sprint-with-us';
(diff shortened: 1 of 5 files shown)


src/back-end/lib/permissions.ts
```diff
@@ -294,12 +298,12 @@ export async function readSWUProposalScore(connection: Connection, session: Sess
...
+export async function createSWUProposal(connection: Connection, session: Session): Promise<boolean> {
+  return isVendor(session) && await hasAcceptedTerms(connection, session);
 }
 
 export async function submitSWUProposal(connection: Connection, session: Session, organization: Organization): Promise<boolean> {
-  return isVendor(session) && !!session && await isUserOwnerOfOrg(connection, session.user, organization.id) && doesOrganizationMeetSWUQualification(organization);
+  return !!session && isVendor(session) && await hasAcceptedTerms(connection, session) && await isUserOwnerOfOrg(connection, session.user, organization.id) && doesOrganizationMeetSWUQualification(organization);
 }
 
 export async function editSWUProposal(connection: Connection, session: Session, proposalId: string, opportunity: SWUOpportunity): Promise<boolean> {
```
dhruvio, line 309: I think we'll need to do the check here. If a vendor is editing a submitted proposal before the proposal deadline, they are technically "resubmitting" a proposal. Applies to both CWU and SWU.

dhruvio (changes requested):
Looks good, one change requested, thanks!

src/back-end/lib/permissions.ts
sutherlanda: Added to editing of proposals as well.

dhruvio (approved):

dhruvio:
Looks good to me

PR #128, opened 2020-09-03
DM-694 strip idpUsername suffix

sutherlanda:
This PR includes a migration to update existing usernames (needs to be tested in dev/test prior to production deployment as the migration would not be easily reversed).

Also includes modification to auth router to remove suffix on new user sign-up.

### DIFF
--- a/src/migrations/knexfile.ts
+++ b/src/migrations/knexfile.ts
@@ -1,16 +1,10 @@
-import dotenv from 'dotenv';
-
-// Working directory changed to src/migrations so ensure env is loaded from correct path.
-dotenv.config({
-  debug: true,
-  path: '../../.env'
-});
+import { DB_MIGRATIONS_TABLE_NAME, POSTGRES_URL } from 'back-end/config';
 
 module.exports = {
   client: 'pg',
-  connection: process.env.POSTGRES_URL,
+  connection: POSTGRES_URL,
   migrations: {
-    tableName: 'migrations',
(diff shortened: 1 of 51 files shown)


src/migrations/knexfile.ts
```diff
@@ -1,10 +1,16 @@
...
+});
 
 module.exports = {
   client: 'pg',
-  connection: POSTGRES_URL,
+  connection: process.env.POSTGRES_URL,
   migrations: {
-    tableName: DB_MIGRATIONS_TABLE_NAME,
+    tableName: 'migrations',
     directory: './tasks'
```
dhruvio, line 14: TODO: I need to implement support for building migrations to JS, and running the JS directly to avoid this change

sutherlanda:
Added an extra check to confirm that when removing `@github` suffix the account is a Vendor type, and similarly for `@idir` and Government account types.  

For the data migration, also allow Admin account types for both `@github` and `@idir`.

dhruvio (commented):
Looks good, I need to fix the migrations build system before we can merge

src/migrations/knexfile.ts
dhruvio: So, I ended up figuring out the simpler solution is to properly resolve the REPOSITORY_ROOT_DIR variable in `back-end/config`. I am doing this by using the `find-up` package to find `package.json` and resolve its dirname. The build system stays the same for migrations.

PR #130, opened 2020-09-23
stabilize IDP auth sign-in flow

sutherlanda:
This PR does the following:

* Adds a database migration which adds a new `idpId` column and transfers values from `idpUsername`
* Modifies the auth router to store users with idpId and retrieve existing users by idpId
* Modifies the shared user model and mocks as needed

### DIFF
--- a/src/back-end/lib/routers/auth.ts
+++ b/src/back-end/lib/routers/auth.ts
@@ -287,13 +287,14 @@ async function establishSessionWithClaims(connection: Connection, request: Reque
   }
 
   let username = getString(claims, 'preferred_username');
+  const idpId = getString(claims, 'idp_id');
 
   // Strip the vendor/gov suffix if present.  We want to match and store the username without suffix.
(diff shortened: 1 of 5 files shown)


src/back-end/lib/routers/auth.ts
```diff
@@ -287,24 +287,26 @@ async function establishSessionWithClaims(connection: Connection, request: Reque
...
     username = username.slice(0, username.lastIndexOf('@'));
   }
 
-  if (!username || !tokenSet.access_token || !tokenSet.refresh_token) {
+  if (!username || !idpId || !tokenSet.access_token || !tokenSet.refresh_token) {
     throw new Error('authentication failure - invalid claims');
   }
 
-  const dbResult = await findOneUserByTypeAndUsername(connection, userType, username);
+  const dbResult = await findOneUserByIdpId(connection, idpId);
```
dhruvio, line 301: Need to query by `idpId` and `userType` because `idpId`s aren't guaranteed to be unique across all users

dhruvio (changes requested):
Looks good overall. One small change requested.

src/back-end/lib/routers/auth.ts
sutherlanda: Right, done. Ty.

PR #132, opened 2020-10-28
DM-29 back-end changes for tracking t&c acceptance

sutherlanda:
* Renamed `acceptedTerms` db column to `acceptedTermsAt`
* Added `lastAcceptedTermsAt` db column.  Uses current value of 'acceptedTerms' for existing users.
* Both columns set to null for newly created users
* Updated shared and back-end libs to use `lastAcceptedTermsAt` in place of `acceptedTerms`
* Permissions for submitting CWU and SWU proposals updated to check for `acceptedTermsAt`
* Front-end pages for profile / legal updated to use `lastAcceptedTermsAt` so that front-end would build, but please review that the changes are correct.


src/back-end/lib/resources/emailNotifications.ts
```diff
@@ -0,0 +1,90 @@
...
+            return invalid({ emailNotification: adt('parseFailure' as const) });
+        }
+      },
+
+      respond: wrapRespond({
+        valid: (async request => {
+          const { session, body } = request.body;
+          switch (body.tag) {
+            case 'updateTerms':
+              handleTermsUpdated(connection);
```
dhruvio, line 77: @sutherlanda Need to set `acceptedTermsAt` to `null` for all users too, so we can track which users haven't yet accepted the latest terms.

dhruvio (changes requested):

src/back-end/lib/resources/emailNotifications.ts
sutherlanda: @dhruvio I had assumed that would be done in whatever method we use to actually update the terms, and that this would just be for a notification.  Do we want that logic in the emailNotification resource, or would it better to have it in a separate resource that we use to update the fixed content? I'm fine with either way, just want to confirm.

sutherlanda: @dhruvio I've made this change in the emailNotifications resource for now so that it doesn't block you.

PR #137, opened 2020-11-06
DM-671 Paginate the Organization listing page

dhruvio:


src/back-end/lib/resources/owned-organization.ts
```diff
@@ -0,0 +1,45 @@
+import * as crud from 'back-end/lib/crud';
```
dhruvio, line 1: @sutherlanda I added a new resource that allows vendors to fetch their owned organizations (readMany only). The front-end needs to fetch these for various pages (e.g. populating list of owned orgs in CWU proposal form). The pagination made this requirement harder to meet with the existing resource.

Question: should I be reading many affiliations to determine owned orgs instead of creating this back-end resource? Or do you think it's OK to have this extra resource?

If you think it's fine to keep this extra resource, could you please review the associated back-end code, and let me know if it looks good to you?

sutherlanda: @dhruvio I think it's fine to have the additional resource.  Affiliations would give you a list of both owned organizations and organizations the user is a member of, so there would be an additional filtering step. Getting owned organizations directly seems like a good idea.  I'll review the back-end code now.

sutherlanda (approved):
LGTM!

PR #138, opened 2020-11-19
Fix a couple typos on Code With Us Opportunity Guide

mark-a-wilson:


realfolk-admin (approved):
Thanks @mark-a-wilson !

PR #171, opened 2021-10-15
Feature/dm 781/docker volume

BCerki:
Addressing ticket DM-781


bdolor:
@BCerki is this ready to be taken out of draft mode? If so, can you also assign a reviewer from the team? Tks!

thegentlemanphysicist:
This works, data persists locally even when I remover the postgres container. I believe having a line in the docs explaining how to wipe the content from the container would be useful

PR #176, opened 2021-10-28
docs: update readme with instructions for using the containerized local environment

BCerki:
This PR updates the readme to explain how to use the containerized environment set up in PR 173.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ cp sample.env .env
 
 Use [Localhost](#localhost) setup if you want to install and run the development environment on your own machine.
 
-Use [Containerized](#containerized) setup if you want to develop within Docker containers using VS Code.
+Use [Containerized](#containerized) setup if you want to develop within Docker containers.
 
 #### Localhost
 
@@ -163,7 +163,7 @@ Ensure the following have been installed on your local machine:
 
 - Docker
 - Docker Compose
-- VS Code Docker extension
+- Optional: VS Code Docker extension
 
 To start the local development environment for the first time, run `docker-compose up` (this builds the containers, starts the database, runs the migrations, and starts the front- and back-ends in watch mode).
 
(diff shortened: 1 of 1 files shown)


wenzowski:
This PR appears to have gone stale. What about [trying crc](https://developers.redhat.com/blog/2019/09/05/red-hat-openshift-4-on-your-laptop-introducing-red-hat-codeready-containers)? That would bring us closer to dev/prod parity.

README.md
```diff
@@ -147,6 +155,42 @@ Then, visit the URL logged to your terminal to view the now locally-running web
...
+
+To start the local development environment for the first time, run `docker-compose up` (this builds the containers, starts the database, runs the migrations, and starts the front- and back-ends in watch mode).
+
+The app will be available at localhost:3000 once you've seen the following among the terminal output:
+
+- `app_1 | Batch 1 run: 49 migrations`
+- `app_1 | [back-end] server started port="3000" host="0.0.0.0"`
+- `app_1 | Running "watch:frontEndStatic" (watch) task` (this is usually the last one to complete, so once you see it, you should be good to go)
+
+To start working within the development environment, open the VS Code Docker extension.
```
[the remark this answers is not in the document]
BCerki: I've added instructions for how to get into the containers via CLI, does that help? It's hard to get away from IDE-specific documentation because the functionality is somewhat IDE-specific--I never created shared volumes because the team decided to use the VS code docker extension and that doesn't require volumes. (Maybe other IDEs also have extensions that don't require volumes?)

I'm totally open to adding shared volumes at some point, but I think it's worth merging this readme into development now because it describes how to use what's already on the development branch.

README.md
```diff
@@ -120,16 +135,9 @@ Once installed, `cd` into this repository's root directory and proceed to instal
...
-```bash
-cp sample.env .env
-# Open and edit .env in your text editor.
-```
-
-Finally, open three terminals and run the following commands:
+Open three terminals and run the following commands:
 
 ```bash
 # Terminal 1
```
thegentlemanphysicist, line 143: I believe we need a note  telling the user that they will need to stop the dm_app container with the command `docker stop dm_app`.  Otherwise `npm run front-end:watch` won't work as expected.

README.md
```diff
@@ -147,6 +155,51 @@ Then, visit the URL logged to your terminal to view the now locally-running web
 
 You can stop the local PostgreSQL container server by running `docker-compose down`. If you wish to completely wipe the container database, including all the data added by the migrations, run `docker volume rm digital_marketplace_dm-vol`.
 
+#### Containerized
```
thegentlemanphysicist, line 158: Okay, after testing locally and realizing that this vs code approach seems to be dependent on the version of VS code being used I have a suggestion.  Since none of us seem to be using docker containers for local app development let's take out all the vs code specific stuff from the readme and just present the Docker commands as a way of running the project locally, but making in clear it's not suitable for app development something like:
"If a local environment needs to be spun up to demo or test the app, the commands `docker-compose build` followed by `docker-compose up`  will build and deploy the containerized app locally. However changes to the codebase will not be reflected in the app until the build command is re-run. This slow turn around makes local development using docker a less desirable approach."

thegentlemanphysicist (changes requested):

README.md
BCerki: Thanks, this seems like a good solution

thegentlemanphysicist (approved):
Looks good

PR #183, opened 2021-11-10
move npm install back to dockerfile

bdolor:
- this moves `npm install` back to the dockerfile
- adds more restrictions to `.dockerignore`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,18 @@
+# local build artifacts
 node_modules
+tmp
+build
+
+# Sonarqube
+.scannerwork
+sonar-project.properties
+
+# Github
+.github
+
+# macOS 
+.DS_Store
+
+# IDE 
+.vscode
+.idea
--- a/src/front-end/Dockerfile
+++ b/src/front-end/Dockerfile
@@ -1,6 +1,10 @@
 FROM --platform=linux/amd64 docker.io/node:10-jessie
 COPY . /usr/app
 WORKDIR /usr/app
-EXPOSE 3000
+RUN npm install
+RUN NODE_ENV=production npm run front-end:build
+RUN npm run back-end:build
 RUN chmod -R 775 /usr/app
+RUN chown -R node:root /usr/app
+EXPOSE 3000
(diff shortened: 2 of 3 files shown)


src/front-end/Dockerfile
```diff
@@ -1,6 +1,10 @@
 FROM --platform=linux/amd64 docker.io/node:10-jessie
 COPY . /usr/app
 WORKDIR /usr/app
-EXPOSE 3000
+RUN npm install
+RUN NODE_ENV=production npm run front-end:build
```
thegentlemanphysicist, line 5: Does this production environment variable mess up local development? i.e. will local development always be prod-like

.dockerignore
```diff
@@ -1 +1,18 @@
-node_modules
\ No newline at end of file
+# local build artifacts
```
thegentlemanphysicist, line 1: Generic question what does the dockerignore do for us?  I guess I'm asking why we need to ignore these local build artifacts?

thegentlemanphysicist (commented):
A couple of questions before I pull this down and start local testing.

src/front-end/Dockerfile
bdolor: Locally, its behaviour suggests it's a red herring. Modifying the input doesn't effect the output with what can be observed inside a running container using both directory size and file composition of the build/front-end directory. Various flavours of input: 

- `NODE_ENV=production`, 
- `NODE_ENV=${NODE_ENV}`
- `NODE_ENV=development` 
- omitting a node environment declaration altogether

produce the same build output: 
image

.dockerignore
bdolor: `dockerignore` prevents local files from being thrown into the image. You can see from the log files of the failed deploy that we could stand to add a few more exclusions (OCP doesn't need git logs to run the app): 

image

The directories 'build', 'tmp' and 'node_modules' are local artifacts, that are created as part of the image build during `RUN` instructions in the Dockerfile. Not unlike local IDE and OS cruft they are artifacts that can also be created locally and thrown into a locally built image.

thegentlemanphysicist (approved):
thanks for making the changes

PR #189, opened 2021-11-20
feat: create build-scan github action

BCerki:
This ticket addresses DM-859. It:

- Runs on merges to development branch
- Builds the app inside a container using Buildah (this is what OpenShift uses for container-building)

Note: Since the action should run only on merges, it doesn't show up with the rest of the CI below since the merge won't happen until after approval. I set it to run once on push so we could see the run: https://github.com/bcgov/digital_marketplace/actions/runs/1483304286. (I couldn't get it to run on [ACT](https://github.com/nektos/act)--its container didn't have Buildah installed)

### DIFF
--- a/.github/workflows/build-scan.yml
+++ b/.github/workflows/build-scan.yml
@@ -2,9 +2,8 @@ name: build-scan
 on: 
   # Since branch protection prevents pushing to development, pushes only happen when merging PRs, so this runs only on merges (there's no way to set GitHub Actions to run on merges)
   push:
-  # temporarily turning off restrictions to test action on github
-    # branches:
-    #   - development
+    branches:
+      - development
 
 jobs:
   build-scan:
--- a/.github/workflows/build-scan.yml
+++ b/.github/workflows/build-scan.yml
@@ -2,8 +2,9 @@ name: build-scan
 on: 
   # Since branch protection prevents pushing to development, pushes only happen when merging PRs, so this runs only on merges (there's no way to set GitHub Actions to run on merges)
   push:
-    branches:
-      - development
+  # temporarily turning off restrictions to test action on github
+    # branches:
+    #   - development
 
 jobs:
   build-scan:
--- a/.github/workflows/build-scan.yml
+++ b/.github/workflows/build-scan.yml
@@ -1,0 +1,19 @@
+name: build-scan
+on: 
+  # Since branch protection prevents pushing to development, pushes only happen when merging PRs, so this runs only on merges (there's no way to set GitHub Actions to run on merges)
+  push:
+    branches:
+      - development
+
+jobs:
+  build-scan:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: redhat-actions/buildah-build@v2
+      with:
+        image: dm-app-image
+        containerfiles: |
+          ../src/front-end/Dockerfile
+
+    
--- a/.github/workflows/build-scan.yml
+++ b/.github/workflows/build-scan.yml
@@ -15,6 +15,6 @@ jobs:
       with:
         image: dm-app-image
         containerfiles: |
-          ../src/front-end/Dockerfile
+          ./src/front-end/Dockerfile
 
     


.github/workflows/build-scan.yml
```diff
@@ -0,0 +1,19 @@
...
+  push:
+    branches:
+      - development
+
+jobs:
+  build-scan:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: redhat-actions/buildah-build@v2
```
bdolor, line 13: What is built depends on where it's going. If a local artifact is we want because we intend to scan it locally (to GH), then this might be an option. Otherwise, we're looking at building/pushing to a container registry.

.github/workflows/build-scan.yml
```diff
@@ -0,0 +1,19 @@
...
+      with:
+        image: dm-app-image
+        containerfiles: |
+          ./src/front-end/Dockerfile
```
bdolor, line 17: Is the omission of the postgres container intentional? Had using docker compose been considered as an alternate way to build both images? Replacing lines 13-17 with the following, for instance.
```
- name: build
run: docker-compose up
```

bdolor (changes requested):
There is a dependency on selecting a container scanning solution.

.github/workflows/build-scan.yml
BCerki: All the container scan actions I've looked at so far involve building the container in one step, and then pushing to a registry is the second step. What I have here is the first step of the push-to-quay.io action: https://github.com/marketplace/actions/push-to-registry

.github/workflows/build-scan.yml
BCerki: I used buildah instead of docker/docker-compose up because docker's not used on openshift and I was aiming for parity. I left out the postgres container because 1) it's not used in prod, 2) it's the same container that's already on Docker Hub.

.github/workflows/build-scan.yml
bdolor: Is `quay.io` is the container scanning solution we're going with?

.github/workflows/build-scan.yml
bdolor: Any benefit to avoiding the assumption that images on Docker Hub are free from vulnerabilities?

BCerki:
> There is a dependency on selecting a container scanning solution.

Shall we put this on hold then?

.github/workflows/build-scan.yml
wenzowski: >  it's the same container that's already on Docker Hub

Careful with that claim, while technically true our `./src/front-end/Dockerfile` here could be updated to [follow best practices](https://cheatsheetseries.owasp.org/cheatsheets/NodeJS_Docker_Cheat_Sheet.html), as [it currently tracks a tag](https://github.com/bcgov/digital_marketplace/blob/084e25444a5c008e807aab9339e70cb42802c54a/src/front-end/Dockerfile#L1) and not a `sha256` hash. As a result there's no guarantee that the docker image we pulled from docker hub using this tag is the same docker image that will be deployed when we rebuild on CI using this tag ([determinism](https://cheatsheetseries.owasp.org/cheatsheets/NodeJS_Docker_Cheat_Sheet.html#1-use-explicit-and-deterministic-docker-base-image-tags)).

.github/workflows/build-scan.yml
wenzowski: There's no reason to limit this project to a single container scanning solution. See [OWASP Docker Security Cheat Sheet RULE #9 - Use static analysis tools](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-9-use-static-analysis-tools) for options to augment [the OCIO's Aqua Service](https://developer.gov.bc.ca/Platform-Services-Security/BC-Government-Aqua-Cloud-Service-Definition).

wenzowski: I think I'm missing something. If we [build with `buildah` we can push with `podman`](https://developers.redhat.com/blog/2019/02/21/podman-and-buildah-for-docker-users#podman_and_container_images) and the output is still an [OCI Image](https://github.com/opencontainers/image-spec). @bdolor how are you imagining having a local copy of the immutable docker layers will be affected by this line and/or how might building locally affect our ability to push layers and tags to a container registry, be that Github, Quay, OCP4, etc.?

bdolor: @wenzowski No reason to assume a limitation has been made, so I gather what's missing for you is the context of this work, which is related to container scanning.

bdolor: @wenzowski the problem to solve is what container scanning solution we're using. The question about local artifacts is trying to discover what the intention was here. Without knowing it was drawing inspiration from a quay.io playbook, it was unclear what this was trying to do.

wenzowski: > Is quay.io is the container scanning solution we're going with?

I understood the context of the work but not the question about Quay–per my comment I wanted to note that a decision to go with quay doesn't preclude use of the OCIO's Aqua Service.

Per rule #9 link above Quay's free engine is called Clair and Acqua's in called Trivy. There are community contributed GH actions for both. Like with git commits, docker layers are immutable objects as validated by checksum.

> What is built depends on where it's going. If a local artifact is we want because we intend to scan it locally (to GH), then this might be an option. Otherwise, we're looking at building/pushing to a container registry.

I understand the gap in clarity and why there would be difficulty understanding intention of the PR due to missing provenance of the approach. My point however was that choice of a scanning solution doesn't depend on where it's going since OCI images can be push/pulled from one registry to another and all–even when using the OpenShift Container Platform’s integrated registry, or an external registry like quay.io–all locations of an image can be used. Per rule #9 the best practice for the build is to be deterministic, so even if we re-build to support a particular scanning technology. However, the best practice would be to [scan images as both part of the CI pipeline, and again as part of the k8s lifecycle](https://sysdig.com/blog/image-scanning-admission-controller/). My understanding of the OCIO Aqua  work is enabling this best practice, thus the comment implying no need to choose between Quay/Clair a Aqua/Trivy. I would advise, however, that in order to best enable a shift left we would want to ensure the deploy-time scan is also run at integration-time, along with any additional integration-time checks and/or technologies enabled.

.github/workflows/build-scan.yml
wenzowski: > Any benefit to avoiding the assumption that images on Docker Hub are free from vulnerabilities?

Supporting this: what if we assume that all images on Docker Hub or the [Red Hat Container Image Registry](https://catalog.redhat.com/software/containers/search) are frequently compromised?

.github/workflows/build-scan.yml
wenzowski: So...if we're introducing build scanning by way of gh actions, I would posit that a "shift left" approach would be to scan twice–once locally so we don't push a bad image, and once when pulling the image so we don't pull down an image that has been compromised since it was scanned on its way into registry, or so we have a guardrail in case of CI breach and first scan isn't trustworthy

SteveChapmanBCDX:
> > There is a dependency on selecting a container scanning solution.
> 
> Shall we put this on hold then?

Yes, let's put it on hold for now.  We'll bring this into the larger CD/CI conversation as we go.

Review on Feb 14, 2022.

VB-at-Bis:
CI/CD pipeline is using Docker, no switch to  buildah and podman (https://developers.redhat.com/blog/2019/02/21/podman-and-buildah-for-docker-users#podman_and_container_images) happened.
Image scanning for security vulnerabilites is performed by Sonar:
https://github.com/bcgov/digital_marketplace/blob/development/.github/workflows/sonar.yml

PR #191, opened 2021-11-22
helm-deploy github action

BryanGK:
This is an initial exploratory version for deploying helm charts to openShift. This is mostly copied from [here](https://github.com/bcgov/security-pipeline-templates/blob/main/.github/workflows/helm-build-deploy.yaml), I would like some feedback on what I have so far. I'm unsure if the GHRC image registry is required in this instance, as well as the secrets. Also, where is it intended that `secrets` or `env` is pulled from?

### DIFF
--- a/.github/workflows/helm-deploy.yml
+++ b/.github/workflows/helm-deploy.yml
@@ -1,0 +1,89 @@
+name: helm-deploy
+on:
+  workflow_run:
+    workflows: ["build-scan"]
+    branches: [development]
+    types: [completed]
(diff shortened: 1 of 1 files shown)


BryanGK:
Ah i see that you're supposed to specify the `env` at the top of the file, so nvm on that question. 👍🏻

bdolor:
PR is in draft mode and there's still some work to do before it's ready for another set of eyes. Removing myself as a reviewer for now.

PR #193, opened 2021-11-23
Dm 900 Fix deployment to prevent override of origin url

thegentlemanphysicist:
Previous deployments were overriding the setting of the origin url (a vanity url that allows keycloak to redirect the app users to the correct url.)  Documentation has been updated,

Test cases
 - Build and deploy development to dev. Confirm that simple auth login works
 - Build and deploy master to test.  Confirm login (both keycloak and simple auth work)

### DIFF
--- a/openshift/templates/app/app-digmkt-deploy.yaml
+++ b/openshift/templates/app/app-digmkt-deploy.yaml
@@ -259,3 +259,4 @@ parameters:
 - name: ORIGIN
   displayName: The vanity url for the deployment
   value: https://${NAME}-${APP_GROUP}-${TAG_NAME}.apps.silver.devops.gov.bc.ca
+  required: true
--- a/openshift/templates/app/app-digmkt-deploy.yaml
+++ b/openshift/templates/app/app-digmkt-deploy.yaml
@@ -18,7 +18,7 @@ objects:
     SERVER_PORT: '3000'
     DATABASE_SERVICE_NAME: ${DATABASE_SERVICE_NAME}-${APP_GROUP}-${TAG_NAME}
     FILE_STORAGE_DIR: "/usr/app"
-    ORIGIN: https://${NAME}-${APP_GROUP}-${TAG_NAME}.apps.silver.devops.gov.bc.ca
+    ORIGIN: ${ORIGIN}
     KEYCLOAK_CLIENT_ID: ${KEYCLOAK_CLIENT_ID}
     KEYCLOAK_URL: ${KEYCLOAK_URL}
     KEYCLOAK_REALM: ${KEYCLOAK_REALM}
@@ -255,3 +255,7 @@ parameters:
 - name: BASIC_AUTH_USERNAME
   displayName: The username for basic auth protection on test environments.
   required: false
+
+- name: ORIGIN
+  displayName: The vanity url for the deployment
+  value: https://${NAME}-${APP_GROUP}-${TAG_NAME}.apps.silver.devops.gov.bc.ca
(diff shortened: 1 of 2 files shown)


BryanGK:
@thegentlemanphysicist have you tested these changes with a deployment?

wenzowski:
> have you tested these changes with a deployment?

Heads up that if we followed the [OWASP Docker Security Cheat Sheet RULE #9 - Use static analysis tools](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-9-use-static-analysis-tools) we'd have a guardrail that would be helping us detect misconfigurations in Kubernetes & Docker when we update a template file. If we're continuously deploying the branch that this PR will merge into we will end up testing these changes with a deployment by the action of merging, so detecting potential misconfigurations early is how we might shift left & raise confidence that by merging this type of PR we'd have shared confidence that it would deploy correctly with the newfound flexibility to override `ORIGIN` while still retaining a good default value.

openshift/templates/app/app-digmkt-deploy.yaml
```diff
@@ -254,4 +254,9 @@ parameters:
...
 - name: BASIC_AUTH_USERNAME
   displayName: The username for basic auth protection on test environments.
-  required: false
\ No newline at end of file
+  required: false
+
+- name: ORIGIN
+  displayName: The vanity url for the deployment
+  value: https://${NAME}-${APP_GROUP}-${TAG_NAME}.apps.silver.devops.gov.bc.ca
+  required: true
```
wenzowski, line 262: Why make the new template variable required when it has a good default value?

wenzowski (commented):
> have you tested these changes with a deployment?

Heads up that if we followed the [OWASP Docker Security Cheat Sheet RULE #9 - Use static analysis tools](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-9-use-static-analysis-tools) we'd have a guardrail that would be helping us detect misconfigurations in Kubernetes & Docker when we update a template file. If we're continuously deploying the branch that this PR will merge into we will end up testing these changes with a deployment by the action of merging, so detecting potential misconfigurations early is how we might shift left & raise confidence that by merging this type of PR we'd have shared confidence that it would deploy correctly with the newfound flexibility to override `ORIGIN` while still retaining a good default value.

openshift/templates/app/app-digmkt-deploy.yaml
thegentlemanphysicist: > @thegentlemanphysicist have you tested these changes with a deployment?

I redeployed dev and test, though it's probably best someone else tries the deployment to make sure the modified readme still work @BryanGK I'll document the two cases on the PR's description

thegentlemanphysicist: @wenzowski I do not believe you can pass in params to a default param in these yaml config files (i tried).  Meaning the ${NAME}, ${APP_GROUP}, and ${TAG_NAME} are treated as strings when the template is processed.  Despite the fact that it's not a 'good' or 'valid' default value the form of the ORIGIN param, having it's for defined will allow us to easily fix the deployment config if the form of these urls ever change or if our pipeline structure changes (i.e. we no longer use a reverse proxy in test, or we have multiple named deployments in dev, etc)

All that said I agree it is a bit _hacky_ I'm not certain the templates are a flexible enough framework to allow something equivalent to a conditional statements to be used.   

ORIGIN: ${ORIGIN ? ORIGIN :  https://${NAME}-${APP_GROUP}-${TAG_NAME}.apps.silver.devops.gov.bc.ca }

wenzowski: Interesting, so the default value parameter isn’t interpolated. That’s a reason to perhaps consider migrating to Helm from the default OpenShift templating system. I agree that an expressive default value like this helps understand what string to pass in to the template.

wenzowski (approved):

BryanGK (approved):
Looks good!

PR #195, opened 2021-11-26
Jest setup

BCerki:
This PR addresses DM-911 installs and configures Jest for front-end testing. It also includes an example unit test suite for the ProgramType component (two tests should pass, one should fail).

Note: I had to use older versions of jest and jest types because the project's typescript version was old. Once Jon finishes the updates, we might be able to go up.

PR #198, opened 2021-11-30
Dm 479 swu checkmark

BCerki:
NOT READY FOR REVIEW
Addresses DM-479

--------------------
DM-479 - Jira
This would be a nice-to-have at this point, but something I think we should consider.
Design decision:  If the team is SWU Qualified, identify with a check mark.

AC:
New “SWU Qualified?” column to Organizations view
If team is SWU Qualified, then check mark appears in “SWU Qualified?” column

--------------------
Notes
- needs to be rebased on development after [jest installation PR](https://github.com/bcgov/digital_marketplace/pull/195) is merged
- needs to finish unit test for `tableBodyRows`
- check with PO about showing an X if the organization isn't SWU qualified--not in the ticket, but would match how it's done in the profile section


SteveChapmanBCDX:
Reviewed with team - closing ticket.  May open JIra ticket DM-479 again in the future.

PR #318, opened 2022-09-01
DM-938: Clean up linters and add prettier

dhruvio:
This PR closes issue: DM-938

Includes tests? N/A
Updated docs? Y

Proposed changes:
- Remove tslint completely and replace with eslint.
- Introduce prettier and an acceptable prettier config.
- Support auto-formatting with prettier in the pre-commit hook.
- Support prettier format-checking in GitHub Action.


.github/workflows/ci.yml
```diff
@@ -13,32 +13,20 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: '16'
-          cache: 'yarn'
```
dhruvio, line 17: I think prettier looks at YAML files too and made this change when I saved it.

bdolor (approved):

PR #328, opened 2022-11-29
DM-1082: Fix how query strings are appended to URLs in front-end CRUD library

dhruvio:
This PR closes issue: DM-1082

Includes tests? N
Updated docs? N

Proposed changes:
- The query string was being inserted to certain URLs in the middle of the path. This PR supports an optional `query` argument for the `make*Action` helper functions to ensure query strings are appended to the end of URLs.
- All relevant front-end HTTP modules have been fixed to use the new `query` parameter.

Additional notes:
- This bug was discovered by @IanFonzie when smoke testing the front-end.

### DIFF
--- a/src/front-end/typescript/lib/http/api/affiliation.ts
+++ b/src/front-end/typescript/lib/http/api/affiliation.ts
@@ -21,8 +21,9 @@ export function readManyForOrganization(
   organizationId: Id
 ): crud.ReadManyAction<Resource.AffiliationMember, string[], unknown> {
   return crud.makeReadManyAction(
-    `${NAMESPACE}?organization=${window.encodeURIComponent(organizationId)}`,
-    (a: Resource.AffiliationMember) => a
+    NAMESPACE,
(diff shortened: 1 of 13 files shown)


sutherlanda:
@dhruvio @bdolor @IanFonzie  I've added fixes for the issues identified by Ian, plus a few others I found incidentally:

1. Fixed an issue where you could not add a team member when creating a SWU proposal
2. Fixed an issue where the Add Team member button would not work for later phases when creating a SWU proposal. For example, if the Prototype phase was required, the Add Team members button for the Implementation phase would do nothing.
3. Fixed an issue where the Draft/Submit buttons were not functioning when completing a proposal. (Due to conflated calls to the backend for fetching proposal and opportunity).
4. Fixed an issue where the confirmation modal for submitting a proposal would not be shown, preventing submission of a proposal.

I've tested that vendors are able to submit CWU and SWU proposals, but I'd recommend we do a bit more testing to make sure we've caught all the edge cases.

PR #333, opened 2022-12-07
Counter Homework Solution

IanFonzie:
Includes tests? N
Updated docs? N

Proposed changes:
- Completed homework for isolating the single counter developed in `docs/learn/front-end/typescript/exercises/02-single-counter/index.tsx` to its own component.

Additional notes:
- Had a bit of trouble with imports where I had to reference the `component` before the function I wanted to use, i.e. `Counter.view`, because `component` was in the module's namespace. I wanted to make the component the default export but I wasn't sure how to type that. I was inspired by the `Table` component, so I might've picked up some bad habits? 
- Also had difficulty determining what the initial state of my "app" component should be. With my current approach, if I used multiple counters they would all need different keys, but that feels wrong? I'm sure I'll get a better idea of how this would work when we move onto the multiple counters exercise.

### DIFF
--- a/docs/learn/front-end/typescript/exercises/02-single-counter/counter.tsx
+++ b/docs/learn/front-end/typescript/exercises/02-single-counter/counter.tsx
@@ -1,0 +1,60 @@
+import { component as component_ } from "front-end/lib/framework";
+import { adt, ADT } from "shared/lib/types";
+import React from "react";
+import { ComponentViewProps } from "front-end/lib/framework/component/base";
+
+/**
+ * Display a number that can be incremented and decremented
+ * by clicking buttons.
+ */
+
+export interface State {
+  counter: number;
+}
+
+type Params = null;
+
+export type Msg = ADT<"increment"> | ADT<"decrement">;
+
+const init: component_.base.Init<null, State, Msg> = () => {
+  return [
+    {
+      counter: 0
+    },
+    []
+  ];
+};
+
+const update: component_.base.Update<State, Msg> = ({ state, msg }) => {
+  switch (msg.tag) {
+    case "increment": {
+      return [state.update("counter", (n) => n + 1), []];
+    }
+    case "decrement": {
+      return [state.update("counter", (n) => n - 1), []];
+    }
+    default:
+      return [state, []];
+  }
+};
+
+type Props = ComponentViewProps<State, Msg>;
+
+const view: component_.base.ComponentView<State, Msg> = ({
+  state,
+  dispatch
+}) => {
+  return (
+    <div>
+      <button onClick={() => dispatch(adt("decrement"))}>- Decrement</button>
+      <div>{state.counter}</div>
+      <button onClick={() => dispatch(adt("increment"))}>+ Increment</button>
+    </div>
+  );
+};
+
+export const component: component_.base.Component<Params, State, Msg, Props> = {
+  init,
+  update,
+  view,
+};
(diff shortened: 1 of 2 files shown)


bdolor:
should the `/exercises` directory be kept clean, but for an  empty`index.tsx` file? 

should this be moved to the `/solutions` directory?

docs/learn/front-end/typescript/exercises/02-single-counter/counter.tsx
```diff
@@ -0,0 +1,60 @@
+import { component as component_ } from "front-end/lib/framework";
+import { adt, ADT } from "shared/lib/types";
+import React from "react";
+import { ComponentViewProps } from "front-end/lib/framework/component/base";
```
dhruvio, line 4: Convention is to namespace framework imports, so this would be `component_.base.ComponentViewProps`. I arbitrarily chose this convention during the refactor, so I'm not personally attached to it. However, I find it keeps things consistent.

docs/learn/front-end/typescript/exercises/02-single-counter/counter.tsx
```diff
@@ -0,0 +1,60 @@
...
+    }
+    case "decrement": {
+      return [state.update("counter", (n) => n - 1), []];
+    }
+    default:
+      return [state, []];
+  }
+};
+
+type Props = ComponentViewProps<State, Msg>;
```
dhruvio, line 41: No need to export this as both `component_.base.ComponentView` and `component_.base.Component` default `Props` to this value.

docs/learn/front-end/typescript/exercises/02-single-counter/counter.tsx
```diff
@@ -0,0 +1,60 @@
...
+      <button onClick={() => dispatch(adt("increment"))}>+ Increment</button>
+    </div>
+  );
+};
+
+export const component: component_.base.Component<Params, State, Msg, Props> = {
+  init,
+  update,
+  view,
+};
```
dhruvio, line 60: To answer your question, you can add the following line to make the component the default export:

```ts
export default component;
```

docs/learn/front-end/typescript/exercises/02-single-counter/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+    case "@newUrl":
+    case "@replaceUrl":
+    default:
+      return [state, []];
+  }
+};
+
+const view: component.base.ComponentView<State, Msg> = ({ state, dispatch }) => {
+  return <Counter.component.view state={state.counter} dispatch={component.base.mapDispatch(dispatch, (msg) =>
+    adt("counter" as const, msg)
```
dhruvio, line 49: I think it's OK to export the individual component functions from the `Counter` module. e.g. `export const view...` so you can simply reference `Counter.view`.

dhruvio (commented):
Great job! Left some very minor suggestions and commentary.

docs/learn/front-end/typescript/exercises/02-single-counter/index.tsx
IanFonzie: Is there a point in exporting the component then?

docs/learn/front-end/typescript/exercises/02-single-counter/counter.tsx
IanFonzie: I think it makes sense to keep it consistent 👍

docs/learn/front-end/typescript/exercises/02-single-counter/counter.tsx
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Display a number that can be incremented and decremented
+ * by clicking buttons.
+ */
+
+export interface State {
+  counter: number;
+}
+
+type Params = null;
```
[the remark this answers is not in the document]
IanFonzie: Interesting, I'll try that. I guess it wouldn't be very often that we have components without params, but if we didn't want them would this be the way to do it?

PR #337, opened 2022-12-14
Dm1100 regression fixes

IanFonzie:
This PR closes issue: DM-1099, DM-1100

Includes tests? N
Updated docs? N

Proposed changes:
- Fetch Proposals and Opportunities on the proposal editing page; fetched two instances of a proposal before.
- Set sidebar state to immutable on the editing page.

Additional notes:
- Regressions found in the big refactor

### DIFF
--- a/src/front-end/typescript/lib/pages/proposal/code-with-us/edit/index.tsx
+++ b/src/front-end/typescript/lib/pages/proposal/code-with-us/edit/index.tsx
@@ -92,7 +92,7 @@ function makeInit<K extends Tab.TabId>(): component_.page.Init<
           immutable({
             proposal: null,
             tab: [tabId, immutable(tabState)],
-            sidebar: sidebarState
+            sidebar: immutable(sidebarState)
           })
         ) as State_<K>,
         [
--- a/src/front-end/typescript/lib/pages/proposal/code-with-us/edit/index.tsx
+++ b/src/front-end/typescript/lib/pages/proposal/code-with-us/edit/index.tsx
@@ -185,21 +185,10 @@ function makeComponent<K extends Tab.TabId>(): component_.page.Component<
               const proposal = proposalResponse.value;
               const opportunity = opportunityResponse.value;
               const affiliations = affiliationsResponse.value;
-              const [sidebarState, sidebarCmds] = Tab.makeSidebarState(
-                tabId,
-                proposal.id,
-                opportunity.id
-              );
               const tabComponent = Tab.idToDefinition(tabId).component;
               return [
-                state
-                  .set("proposal", proposal)
-                  .set("sidebar", immutable(sidebarState)),
+                state.set("proposal", proposal),
                 [
-                  ...component_.cmd.mapMany(
-                    sidebarCmds,
-                    (msg) => adt("sidebar", msg) as Msg
-                  ),
                   component_.cmd.dispatch(
                     component_.page.mapMsg(
                       tabComponent.onInitResponse([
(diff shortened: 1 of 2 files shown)


src/front-end/typescript/lib/pages/proposal/code-with-us/edit/index.tsx
```diff
@@ -185,10 +185,21 @@ function makeComponent<K extends Tab.TabId>(): component_.page.Component<
...
               return [
-                state.set("proposal", proposal),
+                state
+                  .set("proposal", proposal)
+                  .set("sidebar", immutable(sidebarState)),
                 [
+                  ...component_.cmd.mapMany(
+                    sidebarCmds,
+                    (msg) => adt("sidebar", msg) as Msg
+                  ),
```
IanFonzie, line 202: Wasn't sure exactly where I needed to set the immutable `sidebarState`, but this is how the opportunities page does it so I assume it's fine?

src/front-end/typescript/lib/pages/proposal/code-with-us/edit/index.tsx
```diff
@@ -106,8 +106,8 @@ function makeInit<K extends Tab.TabId>(): component_.page.Init<
               proposalId,
               (response) => response
             ),
-            api.proposals.cwu.readOne(opportunityId)(
-              proposalId,
```
IanFonzie, line 110: There was another instance of duplicate proposal requests happening elsewhere in the code, I'm assuming that's also unintended and should be changed?

bdolor: I can't speak to whether it was intentional, but leaving it the way it is doesn't make sense. Would propose the following modification: 

```
 api.opportunities.cwu.readOne(
              opportunityId,
              (response) => response
            ),
```

dhruvio: Yup, it was me, an unintended bug.

src/front-end/typescript/lib/pages/proposal/code-with-us/edit/index.tsx
dhruvio: The sidebar state doesn't need to be re-initialized here. We typically want to re-initialize sidebar state if it requires the result of the async `Cmd`s that make network requests. For the CWU proposal edit page, the `makeSidebarState` function (https://github.com/bcgov/digital_marketplace/blob/d9e48a69290175dfe73edf9f24665d9404496f26/src/front-end/typescript/lib/pages/proposal/code-with-us/edit/tab/index.ts#L95) doesn't need this data.

FYI looks like the bug was caused by `sidebarState` not being made immutable in `init` (see L95: https://github.com/bcgov/digital_marketplace/blob/d9e48a69290175dfe73edf9f24665d9404496f26/src/front-end/typescript/lib/pages/proposal/code-with-us/edit/index.tsx#L95).

dhruvio (changes requested):
One change requested 👍🏽

src/front-end/typescript/lib/pages/proposal/code-with-us/edit/index.tsx
IanFonzie: Yeah, I thought about doing that in L95 for that file; good to know that should patch it up and thanks for the explanation r.e. re-initializing the sidebar.

src/front-end/typescript/lib/pages/proposal/code-with-us/edit/index.tsx
IanFonzie: > I can't speak to whether it was intentional, but leaving it the way it is doesn't make sense. Would propose the following modification:
> 
> ```
>  api.opportunities.cwu.readOne(
>               opportunityId,
>               (response) => response
>             ),
> ```

This ended one ended up being easy to miss, we weren't populating the evaluation criteria for admin/gov users:
Screen Shot 2022-12-14 at 1 52 17 PM

dhruvio (approved):
Nice job 👍🏽

PR #336, opened 2022-12-13
keycloak migration

bdolor:
This PR closes issue: [DM-950]

Includes tests? [N]
Updated docs? [Y]

Proposed changes:
- migrates to new KeyCloak instance

PR #338, opened 2022-12-15
Hotfix - KC migration

bdolor:
This PR closes issue: [DM-950]

Includes tests? [N]
Updated docs? [Y]

Proposed changes:
- changes necessary for gold cluster migration, independent of refactor changes.

Additional notes:
- replaces https://github.com/bcgov/digital_marketplace/pull/336/


src/back-end/lib/routers/auth.ts
```diff
@@ -402,17 +437,18 @@ async function establishSessionWithClaims(
...
-      userType === UserType.Vendor) ||
-    (username.endsWith("@" + GOV_IDP_SUFFIX) &&
-      userType === UserType.Government)
-  ) {
-    username = username.slice(0, username.lastIndexOf("@"));
+  /**
+   * set username - stored in db as idpUsername
+   * set idpId - used as unique identifier in db, MUST be lowercase
+   */
+  let username, idpId: string;
```
IanFonzie, line 444: I don't think it's worth requesting a change, but technically username will have an `any` type as `idpId`'s annotation doesn't apply to it.

IanFonzie (approved):
Looks good, could make one change if you feel like addressing it

src/back-end/lib/routers/auth.ts
bdolor: you're right, single line annotation to combine types doesn't work. thanks! Change made in https://github.com/bcgov/digital_marketplace/pull/338/commits/7e35ffefbffe4ac2f9111e09bff8827ebd3cc3f0

IanFonzie (approved):
These changes seem reasonable. I was slightly concerned about IDIR name concatenation missing either the given name or family name, potentially leading to extra whitespaces. Then I remembered that the government controls that, and they're probably required anyway.

PR #339, opened 2022-12-15
dm1103: fix incorrect button text when a proposal has already been submitted

sutherlanda:
This PR closes issue: [DM-1103]

Proposed changes:
- The http crud api was intending to return a `ResponseValidation` wrapping a `CWUOpportunityProposal`, but it was actually just returning the unwrapped proposal object. This was the root cause of the failed `isValid` check when determining whether an existing proposal was present or not.
- The fix was to ensure the http crud api wrapped the proposal object in a `ResponseValidation`. Validation was already being performed by the `readMany` crud action

### DIFF
--- a/src/front-end/typescript/lib/pages/opportunity/code-with-us/view.tsx
+++ b/src/front-end/typescript/lib/pages/opportunity/code-with-us/view.tsx
@@ -37,7 +37,6 @@ import {
 import { CWUProposalSlim } from "shared/lib/resources/proposal/code-with-us";
 import { isVendor, User, UserType } from "shared/lib/resources/user";
 import { adt, ADT, Id } from "shared/lib/types";
-import { valid } from "shared/lib/validation";
 
 type InfoTab = "details" | "attachments" | "addenda";
 
@@ -57,7 +56,7 @@ export type InnerMsg =
       [
         string,
         api.ResponseValidation<CWUOpportunity, string[]>,
-        api.ResponseValidation<CWUProposalSlim, string[]> | null
+        CWUProposalSlim | null
       ]
     >
   | ADT<"toggleWatch">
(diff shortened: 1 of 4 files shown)


IanFonzie (changes requested):
Looks good, I think we need these changes in SWU as well

sutherlanda:
> Looks good, I think we need these changes in SWU as well

Right, thanks. Added the equivalent fix for SWU as well.

IanFonzie (approved):

src/front-end/typescript/lib/pages/opportunity/code-with-us/view.tsx
```diff
@@ -98,7 +99,7 @@ const init: component_.page.Init<
         viewerUser && isVendor(viewerUser)
           ? api.proposals.cwu.readExistingProposalForOpportunity(
               opportunityId,
-              (response) => response
+              (response) => valid(response)
```
dhruvio, line 102: Checking the type signature for `readExistingProposalForOpportunity`:

https://github.com/bcgov/digital_marketplace/blob/aa017d5d19ba37d3ded0f93827fdf75c88e41cb2/src/front-end/typescript/lib/http/api/proposal/code-with-us.ts#L43-L54

It looks like it handles the validation there already, and passes a `CWUProposalSlim | undefined` to the `handleResponse` function.

In this case, I think the fix is to change the `InnerMsg` type in this file on L59 from:

```ts
api.ResponseValidation<CWUProposalSlim, string[]> | null
```

to:

```ts
CWUProposalSlim | null
```

And update the rest of the module accordingly.

dhruvio: Ah, the use of `unknown` in the HTTP client libraries is really letting a lot of stuff slip through the cracks :( My apologies

dhruvio (changes requested):

dhruvio (approved):

PR #344, opened 2022-12-22
Dm1081 deactivate

bdolor:
This PR closes issue: [issue DM-1081]

Includes tests? [N]
Updated docs? [N]

Proposed changes:
- prevents deactivated users from continuing to receive emails

Additional notes:
- adds relevant copy in modal

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,5 +1,5 @@
 {
-  "version": "1.4.0",
+  "version": "1.3.0",
   "plugins_used": [
     {
       "name": "ArtifactoryDetector"
@@ -205,5 +205,5 @@
       }
     ]
   },
-  "generated_at": "2022-11-28T22:10:37Z"
+  "generated_at": "2022-12-22T00:25:44Z"
 }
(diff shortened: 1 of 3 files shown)


IanFonzie (commented):
This looks good, but I do I wonder, would it be better to set `notificationsOn` to false when we deactivate the user? My thinking is there might be a situation where they want to deactivate but still receive notifications.

Otherwise, we can just stick with this.

bdolor:
> This looks good, but I do I wonder, would it be better to set `notificationsOn` to false when we deactivate the user? My thinking is there might be a situation where they want to deactivate but still receive notifications.

The column `notificationsOn` takes a data type of `timestamp with time zone` and is nullable (same with `deactivatedOn`) so it will only ever hold a date value or nothing. Setting to false is not possible. Deactivating a user account and still receiving notifications is an undesirable behaviour which aligns with an expectation that deactivation severs the relationship fully, eliminating the possibility for a 'zombie' account (dead, yet still alive via emails that just won't quit).

IanFonzie:
> Deactivating a user account and still receiving notifications is an undesirable behaviour which aligns with an expectation that deactivation severs the relationship fully, eliminating the possibility for a 'zombie' account (dead, yet still alive via emails that just

I suppose we could nullify deactivatedOn, but I think you're right in terms of expected behaviour and the UI would be tricky to implement.

Approving.

IanFonzie (approved):

PR #349, opened 2023-01-17
feat: create twu tables

IanFonzie:
Includes tests? [N]
Updated docs? [N]

Proposed changes:
- Create DB tables for TWU based off of latest SWU tables

### DIFF
--- a/src/migrations/tasks/20230116114550_twu_opportunities.ts
+++ b/src/migrations/tasks/20230116114550_twu_opportunities.ts
@@ -1,0 +1,154 @@
+import { makeDomainLogger } from "back-end/lib/logger";
+import { console as consoleAdapter } from "back-end/lib/logger/adapters";
+import Knex from "knex";
+
+const logger = makeDomainLogger(consoleAdapter, "migrations", "development");
+
+export enum TWUOpportunityStatus {
+  Draft = "DRAFT",
+  UnderReview = "UNDER_REVIEW",
+  Published = "PUBLISHED",
+  EvaluationResourceQuestions = "EVAL_QUESTIONS",
+  EvaluationChallenge = "EVAL_C",
+  Awarded = "AWARDED",
+  Suspended = "SUSPENDED",
+  Canceled = "CANCELED"
+}
+
+export enum TWUOpportunityEvent {
+  Edited = "EDITED",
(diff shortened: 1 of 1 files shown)


src/migrations/tasks/20230116114550_twu_opportunities.ts
```diff
@@ -0,0 +1,154 @@
...
+    table.text("location").notNullable();
+    table.specificType("mandatorySkills", "text ARRAY").notNullable();
+    table.specificType("optionalSkills", "text ARRAY").notNullable();
+    table.text("description").notNullable();
+    table.timestamp("proposalDeadline").notNullable();
+    table.timestamp("assignmentDate").notNullable();
+    table.integer("questionsWeight").notNullable();
+    table.integer("challengeWeight").notNullable();
+    table.integer("priceWeight").notNullable();
+    table.enu("serviceArea", Object.values(TWUServiceArea)).notNullable();
```
IanFonzie, line 62: Made the decision to make this a column instead of a table, like swuPhaseCapabilities. This limits us to only one service area per TWU opportunity, but after Brad and I discussed, we determined it would be suitable for our needs.

src/migrations/tasks/20230116114550_twu_opportunities.ts
```diff
@@ -0,0 +1,154 @@
...
+      ["opportunityVersion", "file"],
+      "twuOpportunityAttachments_pkey"
+    );
+    table
+      .uuid("opportunityVersion")
+      .references("id")
+      .inTable("twuOpportunityVersions")
+      .notNullable()
+      .onDelete("CASCADE");
+    table.uuid("file").references("id").inTable("files").notNullable();
```
IanFonzie, line 102: Copied this over from the equivalent SWU table, but it got me thinking, shouldn't we cascade this delete for files?

sutherlanda: If I recall correctly, we don't want to delete from the file table because a hashed file in the database may be used elsewhere (i.e. same file uploaded for multiple opportunities). See `src/back-end/lib/db/file.ts`

sutherlanda (approved):
Looks good

PR #360, opened 2023-01-29
Dm1134 logging with Node memory usage

sutherlanda:
This PR adds configurable logging in production.

Two new configuration parameters were added for the back-end:

`LOG_DEBUG` --> enables debug-level logging in production (per-request)
`LOG_MEM_USAGE` --> enables logging memory usage metrics (also requires debug level logging to be enabeld)

Below is an example of the memory usage logging that will be include with each request log:

```
[request:2f6aaba1-3382-4f28-9a98-38766ee3c5a4] -> GET /images/wayne.jpg sessionId="anonymous"
[request:2f6aaba1-3382-4f28-9a98-38766ee3c5a4] <- 200 1ms
[request:2f6aaba1-3382-4f28-9a98-38766ee3c5a4] memory usage arrayBuffers="0.74 MB" external="2.38 MB" heapUsed="55.83 MB" heapTotal="85.02 MB" rss="147.23 MB"
[request:516acefb-7879-4b39-a203-e5c05e4841dd] -> GET /fonts/BCSans/BCSans-Bold.woff2 sessionId="anonymous"
[request:516acefb-7879-4b39-a203-e5c05e4841dd] <- 200 0ms
[request:516acefb-7879-4b39-a203-e5c05e4841dd] memory usage arrayBuffers="0.95 MB" external="2.6 MB" heapUsed="55.92 MB" heapTotal="85.02 MB" rss="147.28 MB"
[request:b23f2c52-86f9-42d0-9dfc-b31059ab0ee6] -> GET /fonts/BCSans/BCSans-Regular.woff2 sessionId="anonymous"
[request:b23f2c52-86f9-42d0-9dfc-b31059ab0ee6] <- 200 1ms
[request:b23f2c52-86f9-42d0-9dfc-b31059ab0ee6] memory usage arrayBuffers="1.15 MB" external="2.79 MB" heapUsed="55.99 MB" heapTotal="85.02 MB" rss="147.3 MB"
```

[Here](https://nodejs.org/api/process.html#processmemoryusage) is a good explanation of the memory usage metrics.


src/back-end/lib/hooks/logger.ts
```diff
@@ -15,6 +24,19 @@ const hook: RouteHook<unknown, unknown, unknown, unknown, number, Session> = {
...
+      for (const key in heap) {
+        request.logger.debug(
+          `${key} ${
+            Math.round((heap[key as keyof MemoryUsage] / 1024 / 1024) * 100) /
+            100
+          } MB`
+        );
+      }
+      request.logger.debug("---");
+    }
```
dhruvio, line 39: Each logger function (e.g. `logger.debug`) takes a third optional data argument. It's a plain object whose key/value pairs are printed to the console in a single line, delimited by spaces.

e.g.

```ts
logger.debug("domain", "msg", { foo: "bar", baz: 123 });
// [domain] msg foo=bar baz=123
```

It might help simplify the code and reduce log storage to use this feature — you'd only need to call `request.logger.debug("memory usage", heap)` once.

src/back-end/config.ts
```diff
@@ -51,6 +51,12 @@ export const ENV = NODE_ENV;
 
 const logger = makeDomainLogger(consoleAdapter, "back-end:config", ENV);
 
+export const PROD_LOG_DEBUG =
+  parseBooleanEnvironmentVariable(get("PROD_LOG_DEBUG", "0")) ?? false;
```
dhruvio, line 55: I think you could generalize this to `LOG_DEBUG` and have the default value be `ENV === "development" ? "1" : "0"` instead of `"0"`. Then, in `logger/index.ts` you wouldn't have to look at the `ENV` config var, you could only look at `LOG_DEBUG` to turn debug logging on/off.

dhruvio (commented):
Nice. Cool that the stdlib offers memory usage info pretty conveniently via `process`. Left some comments with ideas for your review/discussion.

Also, what do you think about logging the change in memory through the lifecycle of the request. Instead of printing the heap info at the end of the request, we cache the heap info in the `before` hook as `HookState`, then get the heap info again in the `after` hook and subtract the before and after values to figure out if memory is increasing/decreasing through the request. e.g. `rss=+1234 ...`

src/back-end/lib/hooks/logger.ts
dhruvio: Also, forgot to mention, my example is very simplified, I think the megabyte conversion you are doing is quite useful.

src/back-end/config.ts
sutherlanda: Good idea, I'll try that out.

src/back-end/lib/hooks/logger.ts
sutherlanda: So initially, I actually did measure heap usage on the `before` and `after` hooks to get the difference, but I realized that we don't necessarily have to do that. It results in two measurements instead of one, and there wasn't any measurable memory use occurring between requests in my testing. The heap usage at the end of the preceding request was always the exact amount measured at the start of the next request. So if we measure after each request, we're essentially getting the same measurement, with half the calls. I guess the potential disadvantage of doing it this way is that we could potentially miss any heap increases that happen in between requests, but that didn't seem to be happening. I'm happy to revert it back to the initial way I had if you think that's preferable.

I like the suggestion of generalizing using the existing optional parameter for the logging call. I'll make that change.

dhruvio: Ah, sounds good to me, I don't think we need to make the two measurements. Thanks for the explanation.

src/back-end/config.ts
sutherlanda: I've made the suggested changes re: logging formatting and config parameter mentioned above @dhruvio , thanks!

README.md
```diff
@@ -277,7 +277,7 @@ Environment variables that affect the back-end server's functionality are stored
...
 | `TZ`                                    | Time-zone to use for the back-end. Required by the Linux OS that runs the back-end, but not used as application configuration.                                                                                                           |
 | `SHOW_TEST_INDICATOR`                   | A boolean flag (set to `0` for `false`, `1` for `true`) to indicate that an environment is intended for testing purposes (prefixes emails subjects and shows a testing variant of the logo in email notifications). Defaults to `false`. |
-| `PROD_LOG_DEBUG`                        | A boolean flag (set to `0` for `false`, `1` for `true`) to enable debug-level logging in production. Defaults to `false`.                                                                                                                |
+| `LOG_DEBUG`                             | A boolean flag (set to `0` for `false`, `1` for `true`) to enable debug-level logging in production. Defaults to `false`.                                                                                                                |
```
dhruvio, line 280: Might be worthwhile expanding on this to explain how the default value is true for dev and false for prod.

dhruvio (approved):
LGTM 👍🏽

PR #361, opened 2023-01-30
Watch twu opportunities

IanFonzie:
Includes tests? [N]
Updated docs? [N]

Proposed changes:
- Refactors the list page to use helper functions instead of ternary expressions
- Adds TWU opportunities to the list page

### DIFF
--- a/src/front-end/typescript/lib/interfaces/opportunities/index.ts
+++ b/src/front-end/typescript/lib/interfaces/opportunities/index.ts
@@ -1,0 +1,21 @@
+import { Route } from "front-end/lib/app/types";
+import { Id } from "shared/lib/types";
+import { cwu } from 'front-end/lib/interfaces/opportunities/code-with-us';
+import { swu } from 'front-end/lib/interfaces/opportunities/sprint-with-us';
+import { twu } from 'front-end/lib/interfaces/opportunities/team-with-us';
+import { Opportunity } from "front-end/lib/pages/opportunity/list";
+
+export type ListOppHelpers<Opp extends Opportunity['value']> = {
+  getOppViewRoute(id: Id): Route;
+  getOppEditRoute(id: Id): Route;
+  isOpportunityAcceptingProposals(opportunity: Opp): boolean;
+  getOppDollarAmount(opportunity: Opp): number
+}
+
+export function getListOppHelpers(opportunity: Opportunity): ListOppHelpers<Opportunity['value']> {
+  return {
+    cwu,
+    swu,
+    twu
+  }[opportunity.tag];
+}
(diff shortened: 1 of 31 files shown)


IanFonzie:
@dhruvio , you able to take a look at the changes in src/front-end/typescript/lib/interfaces/opportunities/*. I think they're mostly reasonable, maybe a little too specific.

You can ignore anything in src/front-end/typescript/lib/pages/opportunity/team-with-us/view.tsx, I added some nonsense it needed to compile.

src/front-end/typescript/lib/interfaces/opportunities/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  isOpportunityAcceptingProposals(opportunity: Opp): boolean;
+  getOppDollarAmount(opportunity: Opp): number
+}
+
+export function getListOppHelpers(opportunity: Opportunity): ListOppHelpers<Opportunity['value']> {
+  return {
+    cwu,
+    swu,
+    twu
+  }[opportunity.tag];
```
IanFonzie, line 20: This is the the change I'm most concerned about, I'm assuming we want to enumerate all the possible procurement types but this probably isn't the place to do it.

dhruvio: @IanFonzie Great job on the interfaces. My suggestion on module organization:

1. Create an `interfaces/opportunities/types.ts` file. This should contain two types:`ListOppHelpers<Opp ... >` and `OppHelpers<Opp ... >`. The second type should look something like this:

```ts
interface OppHelpers<Opp extends CWUOpportunity | TWUOpportunity | SWUOpportunity> {
  list: ListOppHelpers<Opp>;
}
```

2. Then, in each of `interfaces/opportunities/[code/sprint/team]-with-us.ts`, export one thing:

```ts
import { OppHelpers } from 'front-end/lib/interfaces/opportunities/types';

const helpers: OppHelpers<CWUOpportunity> {
  list: { getOppViewRoute: ..., etc }
  // add more helper groups later on here.
};

export default helpers;
```

3. Finally, in `interfaces/opportunities/index.ts`:

```ts
import cwu from "front-end/lib/interfaces/opportunities/code-with-us";
// etc.

export default { cwu, twu, swu };
```

---

IMO, this approach has 2 benefits:

1. Use standard object syntax to reference opportunity helpers:

```ts
import oppHelpers from "front-end/lib/interfaces/opportunities";

const opp = adt("cwu", ...);
oppHelpers[opp.tag].list.getOppViewRoute();
```

2. Offers you room to add additional interfaces of helpers per opportunity type.

This advice is not meant to prescriptive, ultimately, you should go for the solution that sparks joy 🌟

dhruvio: Also, it may make sense to move the `Opportunity` type into `front-end/lib/interfaces/opportunities/types` for re-use elsewhere.

dhruvio: Also, I think it's fine to abstract the `.tag` away:

```ts
// By doing this...
oppHelpers(opp).list.getOppViewRoute();
// Instead of...
oppHelpers[opp.tag].list.getOppViewRoute();
```

PR #366, opened 2023-02-14
fix: return a string instead of an object

bdolor:
This PR closes issue: [issue DM-1120]

Includes tests? [N]
Updated docs? [N]

Proposed changes:
- the state was returning an object and not the value of the "body" in the object.


IanFonzie (commented):
LGTM

PR #371, opened 2023-02-22
fix: display addenda history in edit tab

bdolor:
This PR closes issue: [issue DM-1121]

Includes tests? [N]
Updated docs? [N]

Proposed changes:
- fixes regression in addenda edit tab where history was not rendering

Additional notes:

**View addenda tab**

image

**Add Addenda state**

image

**Publish new Addenda**

image

### DIFF
--- a/src/front-end/typescript/lib/pages/opportunity/code-with-us/edit/tab/addenda.tsx
+++ b/src/front-end/typescript/lib/pages/opportunity/code-with-us/edit/tab/addenda.tsx
@@ -13,7 +13,6 @@ import { Col, Row } from "reactstrap";
 import { CWUOpportunity } from "shared/lib/resources/opportunity/code-with-us";
 import { adt, ADT } from "shared/lib/types";
 import { invalid, valid } from "shared/lib/validation";
-import { AddendaList } from "front-end/lib/components/addenda";
 
 export interface State extends Tab.Params {
   opportunity: CWUOpportunity | null;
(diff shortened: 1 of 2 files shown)


src/front-end/typescript/lib/pages/opportunity/code-with-us/edit/tab/addenda.tsx
```diff
@@ -97,6 +97,7 @@ const view: component_.page.View<State, InnerMsg, Route> = ({
   dispatch
 }) => {
   if (!state.opportunity || !state.addenda) return null;
+  const existingAddenda = state.opportunity.addenda;
```
dhruvio, line 100: I think you can add the existing addenda to Line 47 of this file when `Addenda.init` is called. That function takes a param `existingAddenda`. This will avoid you needing to use fallback logic in the view function. Same goes for the SWU opportunity.

dhruvio (changes requested):

src/front-end/typescript/lib/pages/opportunity/code-with-us/edit/tab/addenda.tsx
bdolor: complete in https://github.com/bcgov/digital_marketplace/pull/371/commits/38eae5ff0040dbb856274c5967ae37d72e8e94b8

src/front-end/typescript/lib/pages/opportunity/code-with-us/edit/tab/addenda.tsx
```diff
@@ -43,8 +43,9 @@ const update: component_.page.Update<State, InnerMsg, Route> = ({
   switch (msg.tag) {
     case "onInitResponse": {
       const opportunity = msg.value[0];
+      const existingAddenda = opportunity.addenda;
       const [addendaState, addendaCmds] = Addenda.init({
-        existingAddenda: [],
+        existingAddenda: existingAddenda,
```
dhruvio, line 48: FYI you can simply write `existingAddenda,` here instead of the `existingAddenda: existingAddenda`.

dhruvio (approved):

PR #376, opened 2023-03-01
Dm1135 api integration test harness

sutherlanda:
This adds an integration test harness for the API that leverages `jest` and an ephemeral PostgreSQL database.

To run the current integration tests, run `yarn integration-tests:run`. The setup for the environment and launching of the ephemeral database is handled by a bash script `scrips/run-integration-tests.sh`.  This script will launch the dockerized test database, run migrations, export any required environment variables and then launch the integration tests using `jest`.

Since the database is cleared in between tests, a separate schema `digmkt-test` is used as the default for the ephemeral database. This is to eliminate any risk of the tests being run against a non-test database, and clearing out data unintentionally.

The full suite of integrations tests has not yet been added, but this should be useful for supporting the development of TWU as well as any refactors/improvements to existing functionality (i.e. email notifications).

Notes:
* The integration tests are not configured to run in GitHub actions on pull requests, but this can easily be set up if desired.
* In order to prevent a dependency on KeyCloak when running the tests, mock sessions and signed session IDs are used.


bdolor (approved):

.eslintrc.js
```diff
@@ -23,16 +22,15 @@ module.exports = {
     ecmaVersion: 13,
     sourceType: "module"
   },
-  plugins: ["react", "@typescript-eslint", "mocha", "cypress"],
```
bdolor, line 26: Given that Que. has mocha tests already built, is there a more compelling reason to switch to jest?

sutherlanda: Unfortunately, the mocha tests previously built were not in a useable state. They seemed to have been abandoned mid-way through a large refactor. As a result, I was unable to re-use them. They were also designed to make calls to the dev instance of KeyCloak, which is not ideal.

While both test frameworks are good choices, I prefer Jest, especially when it comes to mocking external dependencies. Jest has better performance, and is generally more lightweight, which I think makes it a better choice here IMO. Given the above and that we didn't have a sunken cost into Mocha due to the previous tests being unusable, I opted to use Jest. I am open to switching back if there are advantages I am not aware of though.

PR #381, opened 2023-03-20
Opportunity Data Model Changes

IanFonzie:
This PR closes issue: [DM-1169]

Includes tests? [N]
Updated docs? [N]

Proposed changes:
- Adds twuResources and serviceAreas tables. twuResources are M:1 with twuOpportunityVersions and also have a FK to serviceAreas. Organizations are M:M with serviceAreas and has a corresponding twuOrganizationServiceAreas link table.

Additional notes:
- Most changes are at the DB layer as a temporary stop gap, but we will not need to do any complex data migrations as a result.
- Would appreciate a second set of eyes to look at the migration, I wasn't sure if we needed to handle deletion logic, i.e. if a twuOpportunityVersion gets deleted, should we clear out the corresponding rows in twuOrganizationServiceAreas. I also wasn't sure about enforcing these relationships i.e. a twuOpportunityVersion must have at least one twuResources, but I guess that can be enforced in the application code.

### DIFF
--- a/src/migrations/tasks/20230315112325_service-areas-table.ts
+++ b/src/migrations/tasks/20230315112325_service-areas-table.ts
@@ -1,0 +1,107 @@
+import { makeDomainLogger } from "back-end/lib/logger";
+import { console as consoleAdapter } from "back-end/lib/logger/adapters";
+import Knex from "knex";
+
+const logger = makeDomainLogger(consoleAdapter, "migrations");
+
+enum TWUServiceArea {
+  Developer = "DEVELOPER",
+  DataSpecialist = "DATA_SPECIALIST",
+  ScrumMaster = "SCRUM_MASTER",
+  DevopsSpecialist = "DEVOPS_SPECIALIST"
+}
+
+export async function up(connection: Knex): Promise<void> {
+  // Service Areas
+  await connection.schema.createTable("serviceAreas", (table) => {
+    table.increments("id").primary().unique().notNullable();
+    table.enu("serviceArea", Object.values(TWUServiceArea)).notNullable();
+    table.string("name").notNullable();
+  });
+  Object.keys(TWUServiceArea).forEach(async (key) => {
+    await connection("serviceAreas").insert({
+      serviceArea: TWUServiceArea[key as keyof typeof TWUServiceArea],
+      name: key
+    });
+  });
+  logger.info("Created serviceAreas table");
+
+  // Organization Service Areas
+  await connection.schema.createTable(
+    "twuOrganizationServiceAreas",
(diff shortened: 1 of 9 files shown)


src/migrations/tasks/20230315112325_service-areas-table.ts
```diff
@@ -0,0 +1,107 @@
...
+  );
+  logger.info("Created twuOrganizationServiceAreas table");
+
+  await connection.schema.createTable("twuResources", (table) => {
+    table.uuid("id").primary().unique().notNullable();
+    table.integer("serviceArea").references("id").inTable("serviceAreas");
+    table
+      .uuid("opportunityVersion")
+      .references("id")
+      .inTable("twuOpportunityVersions");
```
bdolor, line 46: I notice opportunities fail on deletion now and wonder if we're missing a delete on cascade somewhere here? I wonder also if the "id" column is necessary? For context, look at how `twuResourceQuestions` is currently doing something similar. Something like the following might work, for instance: 
```  
 table.uuid("opportunityVersion")
      .references("id")
      .inTable("twuOpportunityVersions")
      .notNullable()
      .onDelete("CASCADE");
      table.primary(["opportunityVersion"]);
```

src/migrations/tasks/20230315112325_service-areas-table.ts
```diff
@@ -0,0 +1,107 @@
...
+import Knex from "knex";
+
+const logger = makeDomainLogger(consoleAdapter, "migrations");
+
+enum TWUServiceArea {
+  Developer = "DEVELOPER",
+  DataSpecialist = "DATA_SPECIALIST",
+  ScrumMaster = "SCRUM_MASTER",
+  DevopsSpecialist = "DEVOPS_SPECIALIST"
+}
```
bdolor, line 12: Any good reason why the 'name' column in the ServiceArea table couldn't contain spaces/be user readable? Could flip the keys and values around to accommodate. If we wanted Human Readable labels without string manipulation on the FE, this would be one way to achieve that. 
```
enum TWUServiceArea {
  DEVELOPER = "Developer",
  DATA_SPECIALIST = "Data Specialist",
  SCRUM_MASTER = "Scrum Master",
  DEVOPS_SPECIALIST = "Dev Ops Specialist"
}
```

bdolor (changes requested):
A couple changes request and a comment with respect to the enumerated values for service areas...we're now enumerating values and storing them in the db and so I'm wondering if we want to switch from generating an enum list in the front-end to getting the existing (db) enum from the back-end?

bdolor (approved):

PR #382, opened 2023-03-28
DM-1151: Refactor front-end HTTP modules to improve type safety and inference

dhruvio:
This PR closes issue: DM-1151

Includes tests? N
Updated docs? Y

Proposed changes:

```ts
// Instead of...
api.users.readOne(
  "userId",
  response => adt("gotResponse", response)
) as component_.Cmd<Msg>;

// ... we now do...
api.users.readOne<Msg>()(
  "userId",
  response => adt("gotResponse", response)
);
```

The old iteration (first example above) required type-casting, which can short-circuit the compiler, because it involved casting `unknown` values to `Msg`. However, the new iteration (second example above) uses type parameters to achieve the same outcome without explicitly casting a type. This solution is slightly more cumbersome, though, because it involves wrapping each API function in thunk to the type parameters work as intended.


dhruvio:
This is what the built typedocs look like:

Screen Shot 2023-03-31 at 1 31 12 PM

src/front-end/typescript/lib/http/api/index.ts
```diff
@@ -1,3 +1,42 @@
...
+ * import { User } from "shared/lib/resources/user";
+ *
+ * // Notice the <Msg> concrete type and the extra function call.
+ * const getUserCmd =
+ *   api.users.readOne<Msg>()("userId", response => adt("gotUser", response));
+ * ```
+ *
+ * @module lib/http/api
+ */
+
```
sutherlanda, line 39: This is more out of curiosity, but is there a reason we can't resolve the thunk in the call to create the client function?

For example, for reading a SWU proposal, could we do this? This would avoid having to make the extra function call before using it.

```ts
export function readOne<Msg>(
  opportunityId: Id,
  proposalId: Id,
  handler: HandleResponse<Resource.SWUProposal, string[], Msg> 
): component.cmd.Cmd<Msg> { 
  return crud.makeReadOneAction<RawSWUProposal, Resource.SWUProposal, string[], Msg>(
    NAMESPACE,
    rawSWUProposalToSWUProposal,
    `opportunity=${window.encodeURIComponent(opportunityId)}`
  )(proposalId, handler);
}
```

and then it could be used like this:

```ts
api.proposals.swu.readOne(opportunityId, proposalId, (response) =>
  api.isValid(response) ? response.value : null
),
```

dhruvio: @sutherlanda Neat idea, and yes, that would work. I think the benefit of the thunk approach is that it avoids needing to enumerate the arguments and their type sigs for each client function (`proposalId` and `handler`). However, your suggested solution could result in less boilerplate overall. Do you think it's worthwhile to refactor the code and use this approach instead?

sutherlanda: Ah okay, yeah I'm not sure the benefit of less boilerplate is worth the tradeoff of having to enumerate all the arguments. I don't feel strongly either way, so I think it's good to keep as is if you're good with it.

sutherlanda (approved):

src/front-end/typescript/lib/http/api/index.ts
dhruvio: Same for me, I don't have a strong opinion on the matter either. Will leave it as-is. Thanks!

dhruvio:
Force pushed to rebase off `feature/twu`

IanFonzie (approved):
Sorry for the delay, nothing to add beyond what was already addressed in yours and Andrew's conversation; LGTM.

PR #383, opened 2023-03-29
Twu qualification tab

IanFonzie:
Includes tests? [N]
Updated docs? [N]

Proposed changes:
- https://bcdevex.atlassian.net/browse/DM-1149; basically an interface to qualify orgs for service areas. Anything with a green checkmark in the above ticket is covered in this PR.

Additional notes:
- Looking for feedback on an glaring architectural choices before proceeding with the rest of the items in DM-1149.


src/migrations/tasks/20230327102038_twu-terms.ts
```diff
@@ -0,0 +1,38 @@
+import { generateUuid } from "back-end/lib";
+import { makeDomainLogger } from "back-end/lib/logger";
+import { console as consoleAdapter } from "back-end/lib/logger/adapters";
+import Knex from "knex";
+
+const logger = makeDomainLogger(consoleAdapter, "migrations");
+
+const slug = "team-with-us-terms-and-conditions";
```
bdolor, line 8: this is a duplicate of https://github.com/bcgov/digital_marketplace/pull/377/files#diff-f3b44cb6ba03d1053018cf98a71793369c924049cf8c44d06596098f63140e66R10 

the order of db migrations will require the sequence of merging to be `twu-proposals` followed by `twu-qualifications-tab`. Recommend removing content creation for 'team-with-us-terms-and-conditions' here.

src/front-end/typescript/lib/app/router.ts
```diff
@@ -335,6 +335,17 @@ const router: router_.Router<Route> = {
...
     },
+    {
+      path: prefixPath("/organizations/:id/team-with-us-terms-and-conditions"),
+      makeRoute({ params }) {
+        return {
+          tag: "orgTWUTerms",
+          value: {
+            orgId: params.id || ""
+          }
+        };
```
dhruvio, line 346: I know the other routes use the plain object syntax to make ADTs, but I think it's better to use the `adt` function as it's more concise and provides a better explanation of what is being done. It might be worth addressing this in a small refactor separately.

src/front-end/typescript/lib/index.ts
```diff
@@ -177,3 +179,22 @@ export function getSignInUrl(
...
+export function intersperse<A>(
+  collection: A[],
+  fn: (a: A) => string | ReactElement,
+  separator: string | ReactElement
+) {
+  return map(collection, (val: A, index: number) => [
+    ...(index > 0 ? [separator] : []),
+    fn(val)
+  ]);
+}
```
dhruvio, line 200: It looks like this function is missing a return type signature, and returns a `ReactElement[][]`. The 2D array might be unintentional? If so, you would need to use a `fold` or `reduce` instead of `map`.

I also recommend decoupling this from React by implementing a general `intersperse<A>(collection: A[], separator: A): A[]` in `shared/lib/index.ts` and wrapping that function for a more specific React version in `front-end/lib/index.ts`: `intersperse(collection: Array<string | ReactElement>, separator: string | ReactElement): ReactElement[]`.

In functional programming libraries, the "mapping" functionality is left to `map` or `fmap`, and `intersperse` just focuses on adding the separator to an existing list, hence the omission of the `fn` parameter above, but will leave it with you on how you ultimately want to implement this helper.

src/front-end/typescript/lib/pages/dashboard.tsx
```diff
@@ -305,13 +309,18 @@ const update: component_.page.Update<State, InnerMsg, Route> = updateValid(
...
             const [cwuProposals, swuProposals, organizations] = msg.value.value;
-            const isQualified = organizations.reduce(
+            const isSWUQualified = organizations.reduce(
               (acc, o) => acc || doesOrganizationMeetSWUQualification(o),
               false as boolean
             );
+            const isTWUQualified = organizations.reduce(
+              (acc, o) => acc || doesOrganizationMeetTWUQualification(o),
+              false as boolean
+            );
```
dhruvio, line 319: Opportunity for a helper here: `satisfies<A>(as: A[], fn: (a: A) => boolean): boolean`.

dhruvio (commented):
@IanFonzie Nice work. Left some comments for you to ponder/discuss.

src/back-end/lib/resources/organization.ts
```diff
@@ -345,6 +350,14 @@ const update: crud.Update<
           });
         case "acceptSWUTerms":
           return adt("acceptSWUTerms");
+        case "acceptTWUTerms":
+          return adt("acceptTWUTerms");
+        case "qualifyServiceAreas":
+          if (Array.isArray(value)) {
```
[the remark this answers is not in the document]
IanFonzie: As in use `getStringArray`?

src/front-end/typescript/lib/app/router.ts
IanFonzie: I'll change that one here, but good point about the refactor

src/front-end/typescript/lib/pages/organization/edit/index.tsx
```diff
@@ -188,57 +190,7 @@ function makeComponent<K extends Tab.TabId>(): component_.page.Component<
...
-            }
-            return [];
-          })()
-        },
-        TabbedPage.makeGetParentAlerts(
-          idToDefinition
-        ) as component_.page.Alerts<Msg>
-      );
-    }),
+    getAlerts: getAlertsValid(TabbedPage.makeGetParentAlerts(idToDefinition)),
```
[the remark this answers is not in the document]
IanFonzie: Yup, feature change requested by Steve and Inderdeep. They preferred the alert on their respective qualification pages.

src/front-end/typescript/lib/pages/dashboard.tsx
IanFonzie: I'm pretty sure this functionality could be replicated by Array.prototype.some, i.e. `organizations.some(o => doesOrganizationMeetTWUQualification(o))`. 

Not sure if that still warrants a helper function?

dhruvio: Yup, `some` will do the trick!

src/back-end/lib/db/organization.ts
```diff
@@ -501,3 +533,46 @@ export const readOneOrganizationOwner = tryDb<[Id], User | null>(
     return valid(result ? await rawUserToUser(connection, result) : null);
   }
 );
+
+export const qualifyOrganizationServiceAreas = tryDb<
+  [Id, number[], Session],
```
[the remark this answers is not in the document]
IanFonzie: Changing the type of the PK for serviceAreas ended up being a bit of a time sink because quite a few tables relied on it. I have an idea of what I want to do, but might be better to put it off until after we have more work done; though I want to discuss a migration strategy for it when we get to that point.

I did end up using a more informative ServiceAreaId type which aliases number in the meantime.

src/front-end/typescript/lib/index.tsx
```diff
@@ -180,21 +180,17 @@ export function fileBlobPath(file: Pick<FileRecord, "id">) {
...
-  fn: (a: A) => string | ReactElement,
-  separator: string | ReactElement
-) {
-  return map(collection, (val: A, index: number) => [
-    ...(index > 0 ? [separator] : []),
-    fn(val)
-  ]);
-}
+type IntersperseProps = <A>(props: {
+  collection: A[];
```
dhruvio, line 184: I don't think there is a need for a type parameter `A` here.

`collection: ReactElement[] | string;`

Based on its current usage the TS compiler will automatically coerce the `A` to `unknown` (same problem that we had with the front-end HTTP modules).

Also, I think `IntersperseProps` should be an interface:

```ts
interface IntersperseProps {
  collection: Array<ReactElement | string>;
  separator: ReactElement | string;
}
```

Then, `Intersperse` would be:

```ts
export function Intersperse(props: IntersperseProps): ReactElement {
  // ...
}
```

src/shared/lib/index.ts
```diff
@@ -336,3 +336,22 @@ export function arrayFromRange<Element = number>(
...
+  return reduce(
+    collection,
+    (acc: Array<A | B>, val: A, index: number) => [
+      ...acc,
+      ...(index > 0 ? [separator] : []),
+      val
+    ],
+    []
+  );
+}
```
dhruvio, line 357: I recommend using only one type parameter here: `intersperse<A>(collection: A[], separator: A): A[]`. This keeps the function simple and is more type-safe.

If you need the `A | B` functionality, you can still set the type parameter at the call site:

```ts
const foo = intersperse<ReactElement | string>([<div></div>], "sep");
```

src/migrations/tasks/20230327102038_accept-twu-terms.ts
```diff
@@ -1,30 +1,19 @@
...
-  const now = new Date();
-  const body = "Initial version";
-  const id = generateUuid();
-  await connection("content").insert({ id, createdAt: now, slug, fixed: true });
-  await connection("contentVersions").insert({
-    id: 1,
-    title: slug,
-    body,
-    createdAt: now,
-    contentId: id
```
dhruvio: Is this change intentional?

dhruvio (commented):
Nice changes. Almost there.

src/migrations/tasks/20230327102038_accept-twu-terms.ts
IanFonzie: yup, @bdolor already covered it in a different migration apparently.

src/front-end/typescript/lib/index.tsx
dhruvio: Sorry, one suggestion, you could write `Intersperse` as:

```ts
export const Intersperse: View<IntersperseProps> = ({ collection, separator }) => {
  // ...
}
```

src/shared/lib/index.ts
IanFonzie: K I think I had a misunderstanding about the `Intersperse` API we decided on. Looking at the usage again, what you're suggesting makes sense so I'll change it.

dhruvio (approved):
LGTM, nice job!

PR #387, opened 2023-04-12
Twu glue

IanFonzie:
This PR closes issue: [DM-1160]

Includes tests? [N]
Updated docs? [N]

Proposed changes:
- Conditionally renders start proposal action button and alerts on the view opportunity page depending on vendor's TWU qualification status
- Backend validation to prevent unqualified vendors from submitting proposals on the create proposal page

Additional notes:
- Might want to think about how we handle proponents that participate in a competition but have their qualification status in a service area removed for some reason; not a concern for MVP
- We are rendering the create proposal page for unqualified vendors for both SWU and TWU; we're performing validation to prevent them from submitting but might be worth considering the UX of this after MVP

### DIFF
--- a/src/shared/lib/validation/proposal/team-with-us.ts
+++ b/src/shared/lib/validation/proposal/team-with-us.ts
@@ -2,7 +2,8 @@ import { determineBusinessDays, getNumber, getString } from "shared/lib";
 import {
   MAX_RESOURCE_QUESTION_WORD_LIMIT,
   TWUOpportunity,
-  TWUResourceQuestion
+  TWUResourceQuestion,
+  TWUServiceArea
 } from "shared/lib/resources/opportunity/team-with-us";
 import {
   Organization,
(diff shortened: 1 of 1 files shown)


dhruvio:
Responses to the "additional notes":

> Might want to think about how we handle proponents that participate in a competition but have their qualification status in a service area removed for some reason; not a concern for MVP

We had the same concern for SWU during initial implementation. The guidance from leadership/legal was that qualification status at the time of submission is what matters.

> We are rendering the create proposal page for unqualified vendors for both SWU and TWU; we're performing validation to prevent them from submitting but might be worth considering the UX of this after MVP

Good point. It might be worth replacing the route here with the `"notFound"` page if the vendor is not qualified.

src/front-end/typescript/lib/pages/opportunity/team-with-us/view.tsx
```diff
@@ -152,15 +172,38 @@ const update: component_.page.Update<State, InnerMsg, Route> = ({
...
+                opportunity.serviceArea
+              )
+            ) {
+              return "qualifiedCorrectServiceArea";
+            } else if (qualification !== "qualifiedCorrectServiceArea") {
+              return "qualifiedIncorrectServiceArea";
+            }
+          }
+          return qualification;
+        },
```
dhruvio, line 203: Looks good

dhruvio (approved):
Nice work. Recommending one exceptionally small change, otherwise approved!

PR #389, opened 2023-04-14
Fix/twu proposal regressions

IanFonzie:
This PR closes issue: [DM-1180, DM-1181, DM-1182]

Includes tests? [N]
Updated docs? [N]

Proposed changes:
- Check's the length of team instead of its identity, allowing drafts to be saved without entering any input.
- Validates an organization on a `submit` update to further validate team members and service areas. This is necessary because organizations are optional and we need to pass one to `validateTWUProposalTeam`, as well as the fact that we delete `acceptedTWUProposal` if the sessions belongs to a vendor
- Filter organization's displayed in the create proposal form

Additional notes:
- We'd missed some permissions for readMany TWU proposals, so I enabled those in this PR.

### DIFF
--- a/src/back-end/lib/resources/proposal/team-with-us.ts
+++ b/src/back-end/lib/resources/proposal/team-with-us.ts
@@ -119,13 +119,12 @@ const readMany: crud.ReadMany<Session, db.Connection> = (
       }
 
       if (
-        // TODO - add TWU permissions when ready
-        !permissions.isSignedIn(request.session)
-        // || !(await permissions.readManyTWUProposals(
-        //   connection,
-        //   request.session,
-        //   validatedTWUOpportunity.value
-        // ))
+        !permissions.isSignedIn(request.session) ||
+        !(await permissions.readManyTWUProposals(
+          connection,
+          request.session,
+          validatedTWUOpportunity.value
+        ))
       ) {
         return respond(401, [permissions.ERROR_MESSAGE]);
       }
--- a/src/back-end/lib/resources/proposal/team-with-us.ts
+++ b/src/back-end/lib/resources/proposal/team-with-us.ts
@@ -15,7 +15,7 @@ import {
   validateTWUProposalTeam,
   validateTWUOpportunityId,
   validateTWUProposalId,
-  validateProposalOrganization
+  validateDraftProposalOrganization
 } from "back-end/lib/validation";
 import { get, omit } from "lodash";
 import { getNumber, getString, getStringArray } from "shared/lib";
@@ -571,7 +571,7 @@ const update: crud.Update<
           const { organization, resourceQuestionResponses, attachments, team } =
             request.body.value;
 
-          const validatedOrganization = await validateProposalOrganization(
+          const validatedOrganization = await validateDraftProposalOrganization(
             connection,
             organization,
             request.session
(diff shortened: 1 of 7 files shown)


src/back-end/lib/resources/proposal/team-with-us.ts
```diff
@@ -720,6 +719,20 @@ const update: crud.Update<
...
+            request.session
+          );
+          const organization = getValidValue(validatedOrganization, undefined);
+          if (!organization) {
+            return invalid({
+              proposal: adt("submit" as const, [
+                "An organization must be specified prior to submitting."
+              ])
+            });
+          }
```
IanFonzie, line 734: Not sure if this is the best way to check that an organization exists and the error message is possibly confusing since the user didn't submit an org.

Thoughts?

src/back-end/lib/resources/proposal/team-with-us.ts
```diff
@@ -729,12 +742,17 @@ const update: crud.Update<
...
               await validateTWUProposalTeam(
                 connection,
-                validatedTWUProposal.value.team,
-                request.id
+                validatedTWUProposal.value.team.map(
+                  ({ member, hourlyRate }) => ({
+                    member: member.id,
+                    hourlyRate
+                  })
+                ),
```
IanFonzie, line 750: I wonder if we'd be better off changing the type of the team in `update`? It differs from that of `create`.

src/back-end/lib/resources/proposal/team-with-us.ts
dhruvio: @IanFonzie When submitting a proposal, the entire proposal should be validated, so I think this makes sense. Vendors must select a qualified org to submit a valid proposal, so the error message is helpful (IMO) as it explains this requirement.

However, it looks like `validateProposalOrganization` only checks that the organization exists in the database. It doesn't validate whether or not the organization is TWU qualified. Is this check performed elsewhere?

src/back-end/lib/resources/proposal/team-with-us.ts
dhruvio: Not sure what you mean?

src/front-end/typescript/lib/pages/proposal/team-with-us/lib/components/form.tsx
```diff
@@ -31,7 +31,9 @@ import {
 } from "shared/lib/resources/opportunity/team-with-us";
 import {
   // doesOrganizationMeetTWUQualification,
```
dhruvio, line 33: You can remove this comment

src/back-end/lib/resources/proposal/team-with-us.ts
IanFonzie: Yup, lines 737 -> 757 perform additional validation on the team and provided service areas and 780 should check that they're TWU qualified.

src/back-end/lib/resources/proposal/team-with-us.ts
IanFonzie: eh, kinda hard to explain. i need to do this mapping because of
```
CreateRequestBody {
  opportunity: Id;
  organization?: Id;
  attachments: Id[];
  resourceQuestionResponses: CreateTWUProposalResourceQuestionResponseBody[];
  status: CreateTWUProposalStatus;
  team: CreateTWUTeamMemberBody[];
}
```
but the type of a a stored team member is
```
export interface TWUProposalTeamMember {
  member: UserSlim;
  pending: boolean;
  idpUsername: string;
  hourlyRate: number;
}
```
I think the mapping is fine, but it was difficult to debug because `validateTWUProposalTeam`'s raw argument has type `any`, so I'd like to avoid that headache if the situation comes again (though it might not?)

src/back-end/lib/resources/proposal/team-with-us.ts
dhruvio: Got it. In that case I would use `validateOrganizationId` instead of `validateProposalOrganization` since you need the org to be defined. Is there any other use for `validateProposalOrganization`? If not, you can remove the function from the source code altogether.

src/back-end/lib/resources/proposal/team-with-us.ts
dhruvio: Oh, I see what you mean. Since the difficulty with debugging is caused by the use of `any[]` in `validateTWUProposalTeam`, it might make sense to update that function signature to have a `raw: { member: string, hourlyRate: number }[]` type instead. This would force you to run `getString` and `getNumber` at the call site (as needed), but might result in a more sustainable solution with better type errors.

IanFonzie:
@dhruvio , I'm going to merge. Feel free to approve posthumously or request changes (I'll address them in a different PR if there are any).

PR #390, opened 2023-04-16
Dm1178 export functionality

sutherlanda:
Proposed changes:
- Added front-end links/routes/views for exporting Team With Us proposals
- Included views for viewing all proposals, or individual proposals (same format as SWU)

Additional notes:
- The assigned ticket indicated there should be an 'Export All Proposals' from the anonymized team questions page. Since team questions are instead 'Resource Questions' for TWU, I assumed this 'Export All' link should be added to the anonymized resource questions page. But that does not seem to be implemented yet, or at least not in the same pattern that we have for SWU. If I am mistaken, or there is an implementation for viewing anonymized Resource Questions that I should add this link to, please let me know.


bdolor:
@sutherlanda `twu-evaluation` branch contains the work for viewing anonymized Resource Questions that you're looking for. Does it make more sense to consider re-basing this PR?

dhruvio (approved):
Nice, LGTM. Left one comment about module naming, but that's it. I didn't test the export all page because it requires some manual set-up in the DB. We can test the flows fully once all of the feature work for TWU has been completed.

src/front-end/typescript/lib/pages/proposal/team-with-us/export/all.tsx
```diff
@@ -0,0 +1,224 @@
...
+          ),
+          component_.cmd.andThen(
+            api.proposals.twu.readMany<TWUProposalSlim[]>(opportunityId)(
+              (response) => api.getValidValue(response, [])
+            ),
+            (slimProposals) => {
+              return component_.cmd.map(
+                component_.cmd.sequence(
+                  slimProposals.map(({ id }) =>
+                    api.proposals.swu.readOne(opportunityId)(id, (a) =>
```
bdolor, line 89: update to `twu`?

src/front-end/typescript/lib/pages/proposal/team-with-us/lib/views/edit-tab-header.tsx
```diff
@@ -99,6 +103,19 @@ const ViewTabHeader: component.base.View<Props> = ({
       <Row>
         <Col xs="12">
           <DescriptionList items={items} />
+          <Link
+            newTab
```
bdolor, line 107: Why does this _need_ to open in a new tab?

src/front-end/typescript/lib/pages/proposal/team-with-us/export/all.tsx
sutherlanda: Updated.

src/front-end/typescript/lib/pages/proposal/team-with-us/lib/views/edit-tab-header.tsx
sutherlanda: The ticket indicated style and behaviour of the new link for TWU should be consistent with what was in place for SWU, which opens in a new tab. If that is not the case, then I can update, just let me know.

bdolor: Agreed, the work is consistent with the language of the ticket and should the ticket be free from assumption or error then there is no issue. If we know better, or if there's a different approach that deviates away from preference towards better alignment with WCAG then I guess that's where I'm wondering if there's room to ask if it's necessary.

dhruvio: @bdolor All of the export pages open in a new tab as it was a requirement from government during initial implementation. It's also worth mentioning that a vendor would lose any unsaved changes if they were editing their proposal and clicked the "Export Proposal" link and it didn't open in a new tab.

PR #404, opened 2023-04-30
Dm1164 Add integration tests for user resource

sutherlanda:
This PR changes the following:

- Adds integration tests for the User Resource
- Introduces a new configuration variable called `LOG_LEVEL`. This can be used to control the logging level more accurately. Current log levels are `debug`, `info`, `warn`, `error`, and `none`.  This change was made in this PR since it was required for clearer testing output. When running tests we set the logging level to `error`.
- Adjusts how we query the `RawUser` in the database layer. We specify the fields to select so we don't accidentally include unnecessary fields that are in the db table, but not in the `RawUser` type. This change was required to more accurately validate the output from reading a user via the API in integration testing.
- Added a configuration variable `DISABLE_NOTIFICATIONS` that is enabled when testing to ensure we don't attempt to notify on integration tests.


src/back-end/config.ts
```diff
@@ -54,14 +54,16 @@ export const ENV = NODE_ENV;
 
 const logger = makeDomainLogger(consoleAdapter, "back-end:config");
 
-export const LOG_DEBUG =
-  parseBooleanEnvironmentVariable(
-    get("LOG_DEBUG", ENV === "development" ? "1" : "0")
-  ) ?? false;
+export const LOG_LEVEL =
```
IanFonzie, line 57: I haven't tried, but is it possible to use a type annotation here as a LogLevel so that we don't need to use an assertion later on?

IanFonzie (approved):
This looks good from my perspective; good foundation for the rest of the backend integration tests.

src/back-end/config.ts
IanFonzie, line 57: I haven't tried, but is it possible to use a type annotation here as a LogLevel so that we don't need to use an assertion later on?

sutherlanda: Good idea! I've added this.

dhruvio (approved):
Sorry for the late review, LGTM 👍🏽

PR #407, opened 2023-06-07
Dm717 proposal not awarded status

sutherlanda:
This PR adds integration tests to validate CWU and SWU opportunity and proposal statuses in a scenario where an opportunity with two vendor proposals is closed and one proposal is awarded. This type of testing could be used as a template for testing other scenarios in the app that require extensive setup.

These tests were created during the investigation of the bug outlined in [DM-717].

The integration tests passed, and the bug was not reproducible in the current version of the app.


tests/back-end/integration/resources/code-with-us/status.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  afterEach(async () => {
+    await clearTestDatabase();
+  });
+
+  it("should properly update opportunity and proposal statuses when an opportunity is closed and awarded", async () => {
+    const connection = await connectToDatabase(PG_CONFIG);
+
+    // Setup scenario (create an opportunity with two submitted vendor proposals)
+    const { cwuOpportunity, cwuProposal1, cwuProposal2, testAdminSession } =
+      await setupCWUScenario1();
```
IanFonzie, line 29: Are the scenarios going to be used anywhere else? My only qualm is that I need to hop back and forth between files to gather context about the scenario.

IanFonzie (approved):
Sorry for the delay; I was working on getting the integration tests running in parallel in another PR and it ended up growing up in scope.

tests/back-end/integration/resources/code-with-us/status.test.ts
sutherlanda: I made them thinking they could be useful for testing other functionality. It was sort of modelled after the postman tests, which had to be run in sequence to set things up (like creating a RFQ for an org, inviting team members, etc.). I'm totally open to re-arranging these scenarios or putting them inline though if you find it would be more useful that way.

PR #409, opened 2023-06-14
Refactor/vendor proposal modifications

IanFonzie:
This PR closes issue: [DM-1699]

Includes tests? [N]
Updated docs? [N]

Proposed changes:
- Mostly wording mentioned in DM-1699
- Also adds a new tab to the vendor opportunity view that loads the TWU terms and conditions content

### DIFF
--- a/src/front-end/typescript/lib/pages/opportunity/team-with-us/view.tsx
+++ b/src/front-end/typescript/lib/pages/opportunity/team-with-us/view.tsx
@@ -720,15 +720,15 @@ const KeyDates: component_.base.ComponentView<State, Msg> = ({ state }) => {
               <span className="ml-3">{formatDateAtTime(proposal, true)}</span>
             </p>
             <p className="mb-2">
-              <strong>Contract Award Date</strong>{" "}
+              <strong>Contract Award Date (Anticipated)</strong>{" "}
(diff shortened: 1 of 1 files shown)


src/front-end/typescript/lib/pages/opportunity/team-with-us/view.tsx
```diff
@@ -75,6 +82,7 @@ export type InnerMsg =
         api.ResponseValidation<TWUOpportunity, string[]>,
         TWUProposalSlim | null,
         api.ResponseValidation<Content, string[]>,
+        api.ResponseValidation<Content, string[]>,
```
SteveChapmanBCDX, line 85: Is this change required here?  It looks like a duplication of line 84.

SteveChapmanBCDX (approved):
Everything looks good.  Thanks for getting this done right away.

src/front-end/typescript/lib/pages/opportunity/team-with-us/view.tsx
IanFonzie: yup, it's necessary. We're specifying what our messages should look like and those two happen to share a similar type (because they're both content).

PR #413, opened 2023-07-04
Feat/add backend tests to projects

IanFonzie:
This PR closes issue: [DMM-112]

Includes tests? [Y]
Updated docs? [N]

Proposed changes:
- Adds back-end integration tests to Jest projects
- Update Jest
- Close DB connection when stopping the server
- Catch hook promise failures

Additional notes:
- I still think there's a memory leak in the version of Knex we're using related to [this](https://github.com/Vincit/tarn.js/issues/5). I want to upgrade Knex to a version that doesn't use Bluebird promises after writing a few integration tests 
- There's a bug in this version of Jest that displays an [erroneous warning](https://github.com/jestjs/jest/issues/13576)


src/back-end/index.ts
```diff
@@ -308,5 +314,21 @@ export async function stopServer() {
...
+  try {
+    await new Promise<void>((resolve, reject) => {
+      server.close((error) => {
+        if (error) {
+          reject(error);
+        } else {
+          connectToDatabase()
+            .destroy()
+            .then(() => resolve())
+            .catch(reject);
```
sutherlanda, line 326: Is this required because of the how Knex handles the connections internally? It seems like it would create a new closure (with a new instance of `_connection`) that would immediately be destroyed, but I'm guessing it's being done to force to Knex to clean up?

tests/jest.back-end.integration.ts
```diff
@@ -0,0 +1,13 @@
...
+import unit from "./jest.back-end.unit";
+
+const config: Config.InitialOptions = {
+  ...unit,
+  roots: ["./tests/back-end/integration"],
+  displayName: "back-end:integration",
+  setupFilesAfterEnv: ["<rootDir>/tests/back-end/setup.jest.ts"],
+  globalSetup: "<rootDir>/tests/utils/global-setup.ts",
+  globalTeardown: "<rootDir>/tests/utils/global-teardown.ts"
+};
```
sutherlanda, line 11: It looks like this handles the setup and teardown of the dockerized app for the integration tests. Would it make sense to remove the `scripts/run-integration-tests.sh` script and adjust the `package.json` to directly call jest with this config now?

sutherlanda (approved):

tests/jest.back-end.integration.ts
IanFonzie: Ah yeah, missed that. I'll remove it in my next PR

src/back-end/index.ts
IanFonzie: Yeah, I was a little worried about that too, but when I inspected the running program they connections created _seem_ to reference the same object, i.e. when compared using ===.

PR #442, opened 2023-12-06
Feat/multi user proposal UI

IanFonzie:
This PR closes issue: [DMM-299

Includes tests? [N]
Updated docs? [N]

Proposed changes:
- Poll backend for changes to a "shared" proposal"
- If the incoming updatedAt > state updatedAt, stop polling and display a modal warning the viewer user they can either save their changes offline or continue and overwrite their changes

Additional notes:
- Polling is only active while the viewer user is editing

### DIFF
--- a/src/front-end/typescript/lib/pages/proposal/code-with-us/edit/tab/proposal.tsx
+++ b/src/front-end/typescript/lib/pages/proposal/code-with-us/edit/tab/proposal.tsx
@@ -1,4 +1,4 @@
-import { EMPTY_STRING } from "front-end/config";
+import { EMPTY_STRING, PROPOSAL_POLL_DURATION } from "front-end/config";
 import { makeStartLoading, makeStopLoading } from "front-end/lib";
 import { Route } from "front-end/lib/app/types";
 import * as SubmitProposalTerms from "front-end/lib/components/submit-proposal-terms";
@@ -41,7 +41,8 @@ type ModalId =
   | "saveChangesAndSubmit"
   | "withdrawBeforeDeadline"
   | "withdrawAfterDeadline"
-  | "delete";
+  | "delete"
+  | "orgProposalChangeDetected";
 
 export interface State extends Tab.Params {
   proposal: CWUProposal | null;
@@ -110,7 +111,9 @@ export type InnerMsg =
   | ADT<
       "onDeleteResponse",
       api.ResponseValidation<CWUProposal, DeleteValidationErrors>
-    >;
+    >
+  | ADT<"checkStatus">
+  | ADT<"onCheckStatusResponse", api.ResponseValidation<CWUProposal, string[]>>;
 
 export type Msg = component_.page.Msg<InnerMsg, Route>;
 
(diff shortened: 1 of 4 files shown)


src/front-end/typescript/lib/pages/proposal/code-with-us/edit/tab/proposal.tsx
```diff
@@ -705,6 +713,47 @@ const update: component_.page.Update<State, InnerMsg, Route> = ({
...
+
+      return [
+        state,
+        [
+          component_.cmd.delayedDispatch(
+            PROPOSAL_POLL_DURATION,
+            adt("checkStatus") as Msg
+          )
+        ]
+      ];
```
dhruvio, line 755: nice

src/front-end/typescript/lib/pages/proposal/code-with-us/edit/tab/proposal.tsx
```diff
@@ -705,6 +713,47 @@ const update: component_.page.Update<State, InnerMsg, Route> = ({
...
+              )
+            ]
+          : []
+      ];
+    }
+    case "onCheckStatusResponse": {
+      const incomingProposal = api.getValidValue(msg.value, null);
+      if (!incomingProposal || !state.proposal) {
+        return [state, []];
+      }
```
dhruvio, line 734: If a request randomly fails, you may not want to stop polling altogether. It might be better to keep polling even if you encounter a failure.

src/front-end/typescript/lib/pages/proposal/code-with-us/edit/tab/proposal.tsx
```diff
@@ -933,6 +982,32 @@ export const component: Tab.Component<State, Msg> = {
...
+                <li>
+                  Keep working, and when you submit, your version{" "}
+                  <strong>will</strong> overwrite the previously submitted
+                  changes.
+                </li>
+              </ol>
+            </>
+          ),
+          onCloseMsg: adt("hideModal") as Msg,
+          actions: []
```
dhruvio, line 1009: I suggest adding an "OK" action to simply hide the modal.

src/front-end/typescript/lib/pages/proposal/code-with-us/edit/tab/proposal.tsx
```diff
@@ -705,6 +713,47 @@ const update: component_.page.Update<State, InnerMsg, Route> = ({
         ]
       ];
     }
+    case "checkStatus": {
+      const proposal = state.proposal;
+      return [
+        state,
+        proposal && state.isEditing
```
dhruvio, line 720: I suggest adding a boolean to `state` to indicate if the user has dismissed the `orgProposalChangeDetected` modal.

Once the user has seen and acknowledged the modal, there is no need to show the change detected modal again until they exit form-edit-mode. So, dismissing the modal would set `state.hasSeenChangeDetectedModal = true` (for example).

In this line 720, you could add `&& !state.hasSeenChangeDetectedModal` to the condition to make the polling request to the back-end.

Then, when the user stops editing/saves/cancels, you can reset `state.hasSeenChangeDetectedModal = false` so if the user starts editing again, the polling restarts.

The reason I am making this suggestion is because there may be unbuffered `checkStatus` messages waiting to be evaluated. There is no guarantee that `Cmd`s are executed in order, so handling `Msg`s should try not to make implicit assumptions about which messages will or won't be fired in the future.

dhruvio:
Nice work, @IanFonzie. I think the solution will work well as-is. Added some comments for your consideration.

PR #457, opened 2024-02-06
feat: backend db migration for proposal multi-resources

bdolor:


src/migrations/tasks/20240205135550_twu-proposal-member.ts
```diff
@@ -0,0 +1,84 @@
...
+const logger = makeDomainLogger(consoleAdapter, "migrations");
+
+export async function up(connection: Knex): Promise<void> {
+  /**
+   * Adds a new column to the twuProposalMember table to store the TWUResource
+   * ids
+   */
+  await connection.schema.alterTable("twuProposalMember", function (table) {
+    table
+      .uuid("twuResource")
```
IanFonzie, line 14: Can we call it "resource" since twu is implied? Similar to the existing proposal column.

IanFonzie (approved):

PR #479, opened 2024-05-24
Feat/next previous buttons

IanFonzie:
Includes tests? [N]
Updated docs? [N]

Proposed changes:
- New component for rotating between proposals at tab view
- Currently only applied to team-questions tab
- Proponents are sorted by anonymous proponent name

Additional notes:
- Still need to apply the same changes to TWU

### DIFF
--- a/src/front-end/typescript/lib/pages/proposal/sprint-with-us/lib/views/proposal-tab-carousel.tsx
+++ b/src/front-end/typescript/lib/pages/proposal/sprint-with-us/lib/views/proposal-tab-carousel.tsx
@@ -38,6 +38,16 @@ export function sortProponentsByAnonymousProponentName(
       return compareNumbers(a.order, b.order) * -1;
     });
 }
+
+/**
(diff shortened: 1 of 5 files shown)


src/front-end/typescript/lib/pages/proposal/sprint-with-us/lib/views/proposal-tab-carousel.tsx
```diff
@@ -0,0 +1,114 @@
...
+      }
+    : null;
+
+  const nextRouteParams = proposals[prevIndex]
+    ? {
+        proposalId: proposals[prevIndex].id,
+        opportunityId: proposals[prevIndex].opportunity.id,
+        tab
+      }
+    : null;
```
dhruvio, line 82: @IanFonzie I think `nextIndex` and `prevIndex` usage needs to be switched

dhruvio (changes requested):

PR #486, opened 2024-11-07
feat: submit multiple evaluations draft

IanFonzie:
Rough attempt at validating and updating several question evaluations at once; will point out areas where I have some concerns.

### DIFF
--- a/src/back-end/lib/db/opportunity/sprint-with-us.ts
+++ b/src/back-end/lib/db/opportunity/sprint-with-us.ts
@@ -9,6 +9,7 @@ import {
 import { readOneFileById } from "back-end/lib/db/file";
 import { readOneOrganizationContactEmail } from "back-end/lib/db/organization";
 import {
+  allIndividualSWUTeamQuestionResponseEvaluationsComplete,
   readOneSWUAwardedProposal,
   readSubmittedSWUProposalCount,
   updateSWUProposalStatus
@@ -45,8 +46,7 @@ import {
 } from "shared/lib/resources/proposal/sprint-with-us";
 import {
   SWUTeamQuestionResponseEvaluation,
-  SWUTeamQuestionResponseEvaluationStatus,
-  SWUTeamQuestionResponseEvaluationType
+  SWUTeamQuestionResponseEvaluationStatus
 } from "shared/lib/resources/question-evaluation/sprint-with-us";
 import { AuthenticatedSession, Session } from "shared/lib/resources/session";
 import { User, UserType } from "shared/lib/resources/user";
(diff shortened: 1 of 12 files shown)


src/back-end/lib/db/opportunity/sprint-with-us.ts
```diff
@@ -1652,3 +1664,56 @@ export const addSWUOpportunityNote = tryDb<
...
+          "*"
+        );
+
+      if (!statusRecord) {
+        throw new Error("unable to update team question evaluation");
+      }
+    }
+  });
+
+  // TODO: Check if all individual proposals for a proposal have been submitted;
```
IanFonzie, line 1711: Would this be a reasonable place to do this check? Feels a little messy because the functions in this file are specific to opportunities and we're trying to update proposal statuses.

sutherlanda: Is the check that all team question answers for the proposal have been submitted? I think that's okay. You could put it in the db resource for proposals and invoke it from here to keep it a bit cleaner and avoid this function from getting too long.

src/back-end/lib/resources/opportunity/sprint-with-us.ts
```diff
@@ -1625,6 +1642,101 @@ const update: crud.Update<
             });
           }
         }
+        case "submitIndividualQuestionEvaluations": {
+          for (const id of request.body.value.evaluations) {
```
sutherlanda, line 1646: You could map over this instead since it's synchronous and each validation can be done independently.

src/back-end/lib/db/opportunity/sprint-with-us.ts
```diff
@@ -1671,46 +1676,116 @@ export const submitIndividualQuestionEvaluations = tryDb<
...
+              },
+              "*"
+            );
+
+          if (!statusRecord) {
+            throw new Error("unable to update team question evaluation");
+          }
+
+          // Check if proposal can be moved into question consensus
+          const [
```
sutherlanda, line 1717: I think this makes sense, although you might consider extracting into a separate function (the proposal check piece).

src/back-end/lib/resources/opportunity/sprint-with-us.ts
```diff
@@ -1643,81 +1650,105 @@ const update: crud.Update<
...
         }
         case "submitIndividualQuestionEvaluations": {
-          for (const id of request.body.value.evaluations) {
-            const validatedSWUTeamQuestionResponseEvaluation =
-              await validateSWUTeamQuestionResponseEvaluationId(
-                connection,
-                id,
-                request.session
-              );
+          const validations = await Promise.all(
```
sutherlanda, line 1653: 👍🏻

PR #499, opened 2025-04-04
Feat/delete evaluators or chair

IanFonzie:
Includes tests? [N]
Updated docs? [N]

Proposed changes:
- Support deleting Evaluators from, or the changing the Evaluation Chair on, an opportunity any time before all individual evaluations have been submitted for all of an opportunity's proposals.
- Add "Evaluation Panel" tab to "edit opportunity" page. This tab supports management of panel until the opportunity has consensus submitted.
- Allow an individual Evaluator so edit a draft evaluation for a proposal, but do not allow them to edit an evaluation after submitting it.
- Allow an Evaluation Chair to edit consensus after it has been submitted, but before the opportunity transitions past the question-evaluation phase.

Additional notes:
- Still some statuses for proposals and opportunities to remove; will address those in the migration for backwards compatibility

PR #500, opened 2025-04-04
Refactor/adjust folder structure

IanFonzie:
Same changes as #499 , but reorganizes the evaluation folders to be consistent with the structure we settled on for the DB functions.


dhruvio (approved):
Given this is just a folder/module structure change and it includes the changes of a previously-reviewed PR, l think it's fine to merge.

PR #502, opened 2025-04-08
Feat/allow admins to view consensus

IanFonzie:
Adjust permissions and UI functions to allow admins to view submitted consensuses; includes changes from #499 and #500


src/back-end/lib/permissions.ts
```diff
@@ -901,13 +901,14 @@ export async function readOneSWUTeamQuestionResponseConsensus(
...
-        evaluation.evaluationPanelMember.user.id === session.user.id))
+    (isAdmin(session) ||
+      (isGovernment(session) &&
+        (doesSWUOpportunityStatusAllowGovToViewTeamQuestionResponseEvaluations(
+          evaluation.proposal.opportunity.status
+        ) ||
+          (evaluation.proposal.opportunity.status ===
+            SWUOpportunityStatus.EvaluationTeamQuestionsConsensus &&
+            evaluation.evaluationPanelMember.user.id === session.user.id))))
   );
```
dhruvio, line 912: Looks good. It may be worthwhile just making this easier to read with `if` statements:

```ts
if (!session) return false;

if (isAdmin(session)) return true;

if (isGovernment(session)) {
  if (doesSWUOpportunityStatusAllowGovToViewTeamQuestionResponseEvaluations(evaluation.proposal.opportunity.status))
    return true;
  
  const opportunityInTQConsensus = evaluation.proposal.opportunity.status === SWUOpportunityStatus.EvaluationTeamQuestionsConsensus;
  const sessionIsEvaluator = evaluation.evaluationPanelMember.user.id === session.user.id;
  if (opportunityInTQConsensus && sessionIsEvaluator)
    return true;
}

return false;
```

dhruvio (approved):

src/back-end/lib/permissions.ts
IanFonzie: I'll do this in a different branch; already started making some changes to permissions and I think it'll reduce the number of conflicts

PR #503, opened 2025-04-09
Feat/admin self serve contact list export

alex-struk:
This PR closes issue: [issue #DMM-428]

Includes tests? [N]
Updated docs? [N]

Proposed changes:
- Adds a "Export Contact List" button on the admin's /users path that allows admins to download CSV report of all users in the system through a modal interface that allows for query configuration where user may select from Vendor/Government user types as well as various fields, such as first name, last name, email and organization name


alex-struk:
- I added support for custom endpoints in the backend framework. Previously I was registering the endpoint directly at a high level, but I think the new solution is more graceful as it registers it at the appropriate level (in user.ts). You can see previous solution in this [commit](https://github.com/bcgov/digital_marketplace/pull/503/commits/f7c4ec52baa2e22c0332795f3cbb10acaedde83d)
- I added a new database function to bundle users with organizations in one query. Previously it would pull each affiliation separately for each user which is not ideal. Previous solution is visible in [this commit](https://github.com/bcgov/digital_marketplace/pull/503/commits/3453f5e234a3558f42935a863bece6ddc94b60f6)
- src/front-end/typescript/lib/app/update.ts edit was necessary since it started to throw a typescript error there after related imports
- Front end performance is low when there are a lot of users in the list (interacting with modal is slow). There's a PR from Kaegan where he worked on performance on that page, so I'll re-visit this once I get to that PR.

PR #504, opened 2025-04-11
Refactor/populate evaluations migration

IanFonzie:
Schema changes to allow for backwards compatibility; relevant changes were made to: src/migrations/tasks/20240527213854_add-evaluation-committee-panel-tables.ts and src/migrations/tasks/20240718222006_swu-evaluation-tables.ts

NOTE: swuTeamQuestionResponses still have a score column; I'd prefer to not delete the column on the off chance that we need to migrate back, but was figuring we could rename it to denote that it's deprecated?

### DIFF
--- a/src/migrations/tasks/20240718222006_swu-evaluation-tables.ts
+++ b/src/migrations/tasks/20240718222006_swu-evaluation-tables.ts
@@ -172,6 +172,86 @@ export async function up(connection: Knex): Promise<void> {
   );
   logger.info("Created swuTeamQuestionResponseChairEvaluations table.");
 
+  const now = new Date();
+  const migrationNotes = "MIGRATION_NOTES";
+  const swuTeamQuestionResponsesWithEvaluationPanelMembersQuery = connection(
+    "swuTeamQuestionResponses as responses"
+  )
+    .join("swuProposals as proposals", function () {
+      this.on("responses.proposal", "=", "proposals.id");
+    })
+    .join(
+      connection.raw(
+        "(??) as versions",
+        connection("swuOpportunityVersions")
+          .select("opportunity", "id")
+          .rowNumber("rn", function () {
+            this.orderBy("createdAt", "desc").partitionBy("opportunity");
+          })
+      ),
+      function () {
+        this.on("proposals.opportunity", "=", "versions.opportunity");
+      }
+    )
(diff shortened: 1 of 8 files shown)


src/migrations/tasks/20240718222006_swu-evaluation-tables.ts
```diff
@@ -92,6 +92,46 @@ export async function up(connection: Knex): Promise<void> {
...
+    },
+    []
+  );
+  await connection("swuOpportunityStatuses")
+    .update({ status: SWUOpportunityStatus.EvaluationTeamQuestionsIndividual })
+    .where("id", "in", inProgressStatuses);
+
+  // Maintain deprecated status for historical purposes
+  await connection("swuOpportunityStatuses")
+    .update({ status: SWUOpportunityStatus.DeprecatedEvaluationTeamQuestions })
```
IanFonzie, line 132: @dhruvio I think we might need to change all statuses to SWUOpportunityStatus.EvaluationTeamQuestionsIndividual as well as adding a SWUOpportunityStatus.EvaluationTeamQuestionsConsensus, because this [check](https://github.com/bcgov/digital_marketplace/pull/504/files#diff-3e3de5abd29971dad9a6471f8a13824bda480401b562ac5014ed6d3f63fb237eL102) requires the opportunity to have the status in its history to view them.

IanFonzie: Took a stab at addressing this in https://github.com/bcgov/digital_marketplace/pull/504/commits/1dc501c56feaa5e9a789818037294c43f7d14314; we can revert if necessary.

src/migrations/tasks/20240718222006_swu-evaluation-tables.ts
```diff
@@ -73,6 +73,11 @@ export async function up(connection: Knex): Promise<void> {
     '
   );
 
+  // Unused status; deprecate
+  await connection("swuProposalStatuses")
+    .update({ status: SWUProposalStatus.DeprecatedEvaluatedTeamQuestions })
+    .where({ status: PreviousSWUProposalStatus.EvaluatedTeamQuestions });
+
```
dhruvio, line 80: Looks good

PR #505, opened 2025-04-14
feat: evaluations tab for gov users

alex-struk:
This PR closes issue: [issue #NA]

Includes tests? [N]
Updated docs? [N]

Proposed changes:
- added a tabbed view for the government users on the dashboard
- added a new tab "Evaluations" that shows only opportunities on which the user is on the panel


IanFonzie (commented):
Looks mostly good, the query param logic could be cleaned up a bit.

PR #510, opened 2025-04-28
Feat/generate opportunity report

alex-struk:
This PR closes issue: [issue #DMM-457]

Includes tests? N
Updated docs? N

Proposed changes:
- Allows admin users to generate a complete report for all opportunity types after they are awarded by pressing the "View Complete Competition" button on the summary tab of the opportunity details page


IanFonzie (changes requested):
After further consideration, I don't think reusing the existing components are the best use case for this because:
1. The report is meant to be read-only
2. Components in Marketplace are used for state management
3. We shouldn't be modifying the interface of existing components to pacify the requirements of the new report

Instead the src/front-end/typescript/lib/pages/opportunity/:program/complete/index.tsx pages should should execute the necessary Cmds to load all of the data it needs from the back-end and populate some sort of read-only view; you have a few options for this from least to most preferred:
- Use the existing component's **views** and finesse the data into a shape that conforms to the state it accepts (i.e. component_.page.View<State, InnerMsg, Route>)
- Create new read-only views (i.e. component.base.View\<SomeProps\>)
- Implement a new view for the components used in the report that accepts certain props (not state from the component). Then, refactor the component to "pass through" the relevant props from component state to the view

PR #522, opened 2025-07-11
Feat/generate opportunity report

alex-struk:
This PR closes issue: [issue #DMM-457]

Includes tests? N
Updated docs? N

Proposed changes:
- Allows admin users to generate a complete report for all opportunity types after they are awarded by pressing the "View Complete Competition" button on the summary tab of the opportunity details page


src/front-end/typescript/lib/pages/opportunity/code-with-us/edit/tab/summary.tsx
```diff
@@ -236,5 +245,33 @@ export const component: Tab.Component<State, InnerMsg> = {
...
+    ) {
+      return component_.page.actions.none();
+    }
+    return component_.page.actions.links([
+      {
+        children: "View Complete Competition",
+        symbol_: leftPlacement(iconLinkSymbol("external-link")),
+        button: true,
+        color: "primary" as const,
+        onClick: () => {
```
IanFonzie, line 266: You should be able to use the dest prop with the newTab option instead of this: https://github.com/bcgov/digital_marketplace/blob/development/src/front-end/typescript/lib/views/link.tsx#L114

PR #525, opened 2025-07-23
Feat/system upgrade

alex-struk:
Upgrades node to 22, as well as bootstrap to 5 and other packages


IanFonzie (commented):
Looks mostly good though I had a few questions and comments.

Questions:
- ~Why did we get rid of type: ‘module’ in package.json? Just found it strange considering we're using a .mjs to define the config.~
- Is there any benefit to maintaining grunt for the back-end?

Comments:
- Styles are off
- ~Test indicator wasn’t displayed on dev, despite env variable being set (haven’t looked into it beyond the variable being set and I didn't experience the issue locally)~
- ~We *need* to be able to enable debugging from process.tsx~
- A lot of deprecation warnings

alex-struk:
> Looks mostly good though I had a few questions and comments.
> 
> Questions:
> 
> * ~Why did we get rid of type: ‘module’ in package.json? Just found it strange considering we're using a .mjs to define the config.~
> * Is there any benefit to maintaining grunt for the back-end?
> 
> Comments:
> 
> * Styles are off
> * ~Test indicator wasn’t displayed on dev, despite env variable being set (haven’t looked into it beyond the variable being set and I didn't experience the issue locally)~
> * ~We _need_ to be able to enable debugging from process.tsx~
> * A lot of deprecation warnings

- Renamed eslint.config.mts to js ("module" was set in package.json temporarily, but then removed)
- Completely removed grunt
- Adjusted styles for table headings, table text and buttons paddings. Some minor differences remain around `label` components
- SCSS deprecation warnings cleared, except @import statements (these were added to ignore list). Migration instructs to use `@use` statements, but these must always appear at the top, however bootstrap overrides require importing bootstrap after overrides. Additionally, bootstrap uses @import statements internally, so the usage of these statements follows an established pattern. Once bootstrap upgrades their approach, we can restructure our approach to use new approach using @use/@forward statements.

IanFonzie (approved):
Made my best attempt at matching the styles pre-system upgrade; feel free to take a look or merge.

PR #543, opened 2026-07-10
fix: reject CWU proposal creation as submitted after deadline

alex-struk:
## Problem

A Code With Us proposal could be **created and submitted after the opportunity's deadline**. The `create` endpoint (`POST /api/proposals/code-with-us`) accepted a body with `status: Submitted` and only validated that the opportunity *existed* — it never checked whether the opportunity was still accepting proposals. (The separate draft→submit path is guarded, but the create-directly-as-submitted path was not.)

This was observed in production: a proposal was created ~2 hours after the deadline, after the opportunity had already auto-transitioned to `Evaluation`. Because the close job had already run, the proposal was never moved to `UnderReview` and stayed `Submitted` indefinitely, which:

- is **counted** in the opportunity header (`reporting.numProposals`) but **hidden** from the proposals list (gov/admin don't see `Submitted`), producing a confusing "37 proposals" header vs 36 rows; and
- **blocks the opportunity from advancing** `Evaluation → Processing` (a still-active, unevaluated proposal), so the Award button never appears and the opportunity can't be awarded through the UI.

## Fix

Guard the create handler with `isCWUOpportunityAcceptingProposals(opportunity)` (status `Published` **and** deadline in the future) — the same helper the front-end already uses to gate the create page on load. When `status === Submitted` and the opportunity is no longer accepting proposals, return `400 "This opportunity is no longer accepting proposals."` Draft creation is unaffected.

## Test

Adds an integration test that publishes an opportunity, moves its deadline into the past and closes it (mirroring the production auto-close), then asserts a direct `Submitted` create is rejected.

## Verification

Verified locally end-to-end against the running server + create endpoint:

| Case | Before | After |
|---|---|---|
| Open opportunity → submit (control) | 201 | 201 |
| Closed/past-deadline opportunity → submit | 201 (bug) | 400 |

`back-end:typecheck` passes.

## Follow-up (not in this PR)

- Defense-in-depth: apply the same guard to the `update` submit action (covers a cancelled-but-not-yet-expired opportunity), deferred by request.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/back-end/lib/resources/proposal/code-with-us.ts
+++ b/src/back-end/lib/resources/proposal/code-with-us.ts
@@ -19,6 +19,7 @@ import {
 import { get, omit } from "lodash";
 import { getNumber, getString, getStringArray } from "shared/lib";
 import { FileRecord } from "shared/lib/resources/file";
+import { isCWUOpportunityAcceptingProposals } from "shared/lib/resources/opportunity/code-with-us";
 import {
   createBlankIndividualProponent,
   CreateCWUProposalStatus,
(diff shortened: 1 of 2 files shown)


SteveChapmanBCDX (approved):
Reviewed and approved.

